PR #4, opened 2019-03-18
adding component to show on unresolved config state

andrewkikot:


src/assets/scss/xm/parts/_base.scss
```diff
@@ -82,6 +82,10 @@ body {
   color: rgba(0, 0, 0, 0.26) !important;
   cursor: not-allowed !important;
 
+  > xm-loader {
```
sergeysenja1992: '>'  это не опечатка?

andrewkikot: No, it is not)

PR #52, opened 2019-05-28
Feature. Link-detail-search-section. Added query from UI spec config …

ozelenyi:
…(filterQuery) for links list additional filtering.


sergeysenja1992:
Maybe better to move this logic to lep?

ozelenyi:
No moving to lep

PR #68, opened 2019-07-08
Add redirect after function call if location header is returned.

staciedev:


major13ua:
New function feature: do router redirect to other screen or widget.

src/app/xm-entity/function-call-dialog/function-call-dialog.component.ts
```diff
@@ -12,6 +12,7 @@ import { XmEntity } from '../shared/xm-entity.model';
 import { getFileNameFromResponseContentDisposition, saveFile } from '../../shared/helpers/file-download-helper';
 import { BehaviorSubject, merge, Observable, of } from 'rxjs';
 import { catchError, filter, finalize, share, tap } from 'rxjs/operators';
+import {Router} from '@angular/router';
```
andrewkikot: import sapces

PR #92, opened 2019-08-09
Disable Cancel button when attachment is loading.

staciedev:


src/app/xm-entity/attachment-detail-dialog/attachment-detail-dialog.component.html
```diff
@@ -81,7 +81,7 @@ <h4 class="modal-title text-left">
     </div>
   </div>
   <div class="modal-footer">
-    <button mat-raised-button type="button" class="btn mr-1" data-dismiss="modal" (click)="onCancel()">
+    <button mat-raised-button type="button" class="btn mr-1" data-dismiss="modal" (click)="onCancel()" [disabled]="showLoader">
```
sergeysenja1992: but if I want to cancel uploading?

staciedev: You cannot cancel a request already sent. To cancel uploading you will need to delete the file, but there is other functionality for deleting. This button was intended just to close the window, as I understand.

PR #100, opened 2019-08-15
Bugfix: on list modification event, reload currently active tab

staciedev:


src/app/ext-commons/ext-common-entity/entity-fab-actions/entity-fab-actions.component.ts
```diff
@@ -112,7 +112,6 @@ export class EntityFabActionsComponent implements OnInit, OnDestroy {
                     : this.navigate(path, {});
                 return;
             }
-            this.eventManager.broadcast({name: XM_EVENT_LIST.XM_ENTITY_LIST_MODIFICATION});
```
sergeysenja1992: why?

PR #102, opened 2019-08-15
adding possibility to show user description in sidebar

andrewkikot:


src/app/layouts/sidebar/sidebar.component.html
```diff
@@ -23,6 +23,11 @@
       <a data-toggle="collapse" href="#collapseExample" class="collapsed">
         <span class="user-name">
             <em class="truncated pr-4">{{principal.getName()}}</em>
+            <em class="user-description text-muted truncated pr-4"
+                *ngIf="(userDescription$ | async)"
```
major13ua: you can use only 1 async - (userDescription$ | async as varName) and then just use varName

src/app/layouts/sidebar/sidebar.component.ts
```diff
@@ -399,6 +409,22 @@ export class SidebarComponent implements OnInit, OnDestroy, AfterViewInit {
         }
     }
 
+    private getUserDescription(config: any): Observable<any> {
```
major13ua: ctachError and return {} in any error, to prevent async close

andrewkikot: +

src/app/layouts/sidebar/sidebar.component.html
andrewkikot: +

PR #103, opened 2019-08-16
Make possible to turn off sorting in entity list components via config.

staciedev:


src/app/xm-entity/entity-list-card/entity-list-card-options.model.ts
```diff
@@ -39,7 +39,8 @@ export class FieldOptions {
                 public func?: string,
                 public action?: ActionOptions,
                 public actions?: ActionsOptions[],
-                public actionsListPrivileges?: string[]) {
+                public actionsListPrivileges?: string[],
+                public sortingOff?: boolean) {
```
major13ua: sortable?: boolean

PR #107, opened 2019-08-21
Switch from GET to POST request

staciedev:


src/app/shared/jsf-extention/widgets/ext-query-select/ext-query-select.component.ts
```diff
@@ -119,7 +119,7 @@ export class ExtQuerySelectComponent implements OnInit, OnDestroy {
     }
 
     private fetchOptions(query: any): Observable<ISelectOption[]> {
-        return this.http.get(this.settings.url, {params: query})
+        return this.http.post(this.settings.url, query)
```
major13ua: please add exception handler

staciedev: done

PR #147, opened 2019-11-05
Core. Entity-list-card. Hide pagination when insufficient entities

ozelenyi:


ozelenyi:
https://tracker.jbs.com.ua/issues/21033

PR #149, opened 2019-11-06
Link-detail-search-section. Replaced link spec icon with entity spec icon

ozelenyi:
for the entity's links type "SEARCH" in the list of entities selection
AS IS: the icons shown in the list refers to the icon of the link type or parent entity type (PRICES)
TO BE: icons shown in the list should be taken from the linked entity full type (PRICES.USAGE)

### DIFF
--- a/src/app/shared/helpers/xm-entity-icon.pipe.ts
+++ b/src/app/shared/helpers/xm-entity-icon.pipe.ts
@@ -1,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Spec } from '../../xm-entity';
+
+@Pipe({name: 'xmEntityIcon'})
+export class XmEntityIconPipe implements PipeTransform {
+
+  public transform(typeKey: string, spec: Spec): string {
+      return spec.types.filter( (t) => t.key === typeKey).shift().icon;
+  }
+
+}
--- a/src/app/shared/helpers/xm-entity-icon.pipe.ts
+++ b/src/app/shared/helpers/xm-entity-icon.pipe.ts
@@ -5,8 +5,9 @@ import { Spec } from '../../xm-entity';
 export class XmEntityIconPipe implements PipeTransform {
 
   public transform(typeKey: string, spec: Spec): string {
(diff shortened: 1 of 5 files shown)


src/app/xm-entity/link-detail-dialog/link-detail-search-section.component.html
```diff
@@ -30,7 +30,7 @@
         <td>
           <div class="xm-avatar-img-container">
             <img src="{{xmEntity.avatarUrl ? xmEntity.avatarUrl : './assets/img/placeholder.png'}}">
-            <i class="material-icons" *ngIf="linkSpec.icon && !xmEntity.avatarUrl">{{linkSpec.icon}}</i>
+            <i class="material-icons" *ngIf="!xmEntity.avatarUrl">{{getEntitySpec(xmEntity.typeKey).icon}}</i>
```
andrewkikot: Maybe we should use helper pipe to get icon?

src/app/shared/helpers/xm-entity-icon.pipe.ts
```diff
@@ -0,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Spec } from '../../xm-entity';
+
+@Pipe({name: 'xmEntityIcon'})
+export class XmEntityIconPipe implements PipeTransform {
+
+  public transform(typeKey: string, spec: Spec): string {
+      return spec.types.filter( (t) => t.key === typeKey).shift().icon;
+  }
```
andrewkikot: Wouldn`t it fail if filter func do not find coincidence?

src/app/shared/shared.module.ts
```diff
@@ -143,6 +144,7 @@ const MODULES = [
         FileUploadComponent,
         PrivacyAndTermsDialogComponent,
         PIPES,
+        XmEntityIconPipe,
```
andrewkikot: maybe move into PIPE const?

src/app/shared/helpers/xm-entity-icon.pipe.ts
ozelenyi: Agree. How about this `return entitySpec.icon || '';`

PR #160, opened 2019-11-26
Task/21913 Give ability to hide ALL option in fast search

ozelenyi:


src/app/xm-entity/entity-list-card/entity-list-card.component.ts
```diff
@@ -357,7 +238,174 @@ export class EntityListCardComponent implements OnInit, OnChanges, OnDestroy {
         });
     }
 
-    private alert(type, key) {
+    private getEntitiesUIConfig(): void {
+        this.xmConfigService.getUiConfig().pipe(
```
sergeysenja1992: ui config it's not correct place for it

PR #182, opened 2020-01-09
feat(account): add logout page

Kaskyi:


src/app/account/logout/logout.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    public ngOnInit(): void {
+        this.logout();
+    }
+
+    public logout(): void {
+        this.loginService.logout();
+        this.xmEntitySpecWrapperService.clear();
+        this.router.navigate(['']);
+        /** TODO: transform below as a listener of auth state */
+        const body = document.querySelector('body');
```
Zabavsky: %)

src/app/account/logout/logout.component.html
```diff
@@ -0,0 +1 @@
+<xm-loader [showLoader]="true"></xm-loader>
```
Zabavsky: Why you use here boolean constant instead of something more dynamic?

Kaskyi: I think, for the logout page we need some friendly design. In the current solution its infinite loader

Kaskyi:
this branch was fixed and merged into developer

PR #185, opened 2020-01-14
deps: update minor deps

Kaskyi:


Kaskyi:
In case if u have a problem with tslint, Make sure that u use node v10.16.2

PR #206, opened 2020-02-11
Develop request cache

Kaskyi:


src/app/shared/core/src/cache/request-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+    get(): Observable<T | null>;
+
+    forceReload(): void;
+
+    ngOnDestroy(): void;
+
+    clear(): void;
+}
+
+export class RequestCache<T> implements IRequestCache<T> {
```
didyk: We have powerful DI in Angular. I think would be better use DI instead of 'new RequestCache()'

PR #216, opened 2020-02-18
Develop version

Kaskyi:


src/app/modules/xm-ribbon/xm-ribbon.component.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    public show$: Observable<boolean>;
+    public serverEnv: string = getServerEnvironment();
+
+    constructor(protected uiConfigService: XmUiConfigService<{ ribbon: boolean }>) {
+    }
+
+    public ngOnInit(): void {
+        this.show$ = this.uiConfigService.cache$.pipe(
+            filter((i) => Boolean(i)),
```
didyk: The object is always true

PR #225, opened 2020-02-21
Feature/angular 9

Kaskyi:


src/polyfills.ts
```diff
@@ -1,3 +1,7 @@
+/***************************************************************************************************
+ * Load `$localize` onto the global scope - used if i18n tags appear in Angular templates.
+ */
+import '@angular/localize/init';
```
Kaskyi: Could be removed after `ng-jhipster` be updated(or excluded)

PR #228, opened 2020-02-25
Develop alert(toaster) service

Kaskyi:


src/app/shared/toaster/xm-toaster.service.ts
```diff
@@ -0,0 +1,71 @@
...
+                params.close = (args) => {
+                    params.close(args);
+                    observer.next(args);
+                    observer.complete();
+                };
+            } else {
+                params.close = close;
+            }
+
+            this.alertService.addAlert(params as JhiAlert, []);
```
Sumragen: Do we really need to refer on `JhiAlert` interface?

Kaskyi: We use the `Partial <JhiAlert>`, and it is not combined with `ToasterConfig`

Issue #231, opened 2020-02-25
Visual issues on gateway and User management (develop)


amedvedchuk:
1. On Gateway page status icons are not aligned by left side:
![image](https://user-images.githubusercontent.com/10553379/75251012-6364d780-57e2-11ea-9239-153b74a101f9.png)
2. On User management screen Add button rendered incorrectly:
![image](https://user-images.githubusercontent.com/10553379/75251096-94dda300-57e2-11ea-90d9-395fc74fe616.png)

PR #243, opened 2020-02-28
fix: mat-fab color #231

Kaskyi:
#231

### DIFF
--- a/src/app/admin/gateway/gateway.component.html
+++ b/src/app/admin/gateway/gateway.component.html
@@ -34,8 +34,8 @@
                 <div class="table-responsive">
                   <table class="table table-striped" *ngIf="route">
                     <tr *ngFor="let instance of route.serviceInstances">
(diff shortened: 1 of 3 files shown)


src/app/admin/gateway/gateway.component.html
```diff
@@ -34,8 +34,8 @@ <h4 class="card-title" jhiTranslate="gateway.title">Gateway</h4>
                 <div class="table-responsive">
                   <table class="table table-striped" *ngIf="route">
                     <tr *ngFor="let instance of route.serviceInstances">
-                      <td><a href="{{instance.uri}}" target="_blank">{{instance.uri}}</a></td>
-                      <td>
+                      <td [style.width]="'20%'"><a href="{{instance.uri}}" target="_blank">{{instance.uri}}</a></td>
```
Sumragen: Is it nice to set inline style?

Kaskyi: Depends, in this case its OK

PR #248, opened 2020-03-04
Develop alert refactor

Kaskyi:


src/app/xm-entity/calendar-card/calendar-card.component.ts
```diff
@@ -49,21 +50,21 @@ export class CalendarCardComponent implements OnChanges {
...
+        this.alertService.open({
+            title: 'xm-entity.calendar-card.delete.title',
             showCancelButton: true,
             buttonsStyling: false,
             confirmButtonClass: 'btn mat-button btn-primary',
             cancelButtonClass: 'btn mat-button',
-            confirmButtonText: this.translateService.instant('xm-entity.calendar-card.delete.button'),
-        }).then((result) => {
+            confirmButtonText: 'xm-entity.calendar-card.delete.button',
+        }).subscribe((result) => {
```
Sumragen: Probably we could use `pipe`, like
```
.pipe(
  filter(Boolean),
  switchMap(() => this.eventService.delete(event.id)
)
```

Sumragen: Then be sure that we mark this as a tech debt note or task.

PR #256, opened 2020-03-06
feat(toaster): add options

Kaskyi:


src/app/shared/toaster/xm-toaster.service.ts
```diff
@@ -1,10 +1,15 @@
 import { Injectable } from '@angular/core';
 import { TranslateService } from '@ngx-translate/core';
+import * as _ from 'lodash';
 import { JhiAlert, JhiAlertService } from 'ng-jhipster';
 import { Observable } from 'rxjs';
 
 export interface ToasterConfig extends Partial<JhiAlert> {
     text?: string;
```
Sumragen: What is the difference between `text` and `msg` property?

src/app/shared/toaster/xm-toaster.service.ts
```diff
@@ -24,7 +29,10 @@ export class XmToasterService {
         }
 
         if (params.msg) {
-            params.msg = this.translateService.instant(params.msg);
+            const opts = params.textOptions || {};
```
Sumragen: Let's just say what we can see.
"In order to specify translation options of field `msg` I need to use `textOptions` property. By the way, we also have property `text` which are override `msg`."
That will be pretty clear for developers when they are just check under the hood of that service, but interface isn't so clear at first sight.
My main concern is semantic thing of it.

PR #272, opened 2020-03-09
feat: add thow error to utils

Kaskyi:


src/app/modules/xm-shared-operators/src/throw-error.ts
```diff
@@ -0,0 +1,9 @@
+import { environment } from '@xm-ngx/environment';
+
+export function throwError(text: string): never {
```
Sumragen: Could it be an expected behavior?

Kaskyi: yes, for those cases this function does not work. You can use it when u need to show assertion Errors

PR #282, opened 2020-03-13
feat(permission): improve doc

Kaskyi:


src/app/shared/privilege/permission.directive.ts
```diff
@@ -63,10 +66,11 @@ export class PermissionDirective {
...
 
     @Input()
-    public set xmPermission(value: string | string[]) {
-        if (!value) {
+    public set xmPermission(value: string | boolean | string[]) {
+        if (!value || value === true || value.length === 0) {
             // Show by default
-            this.context.allow = true;
+            this.context.$implicit = [];
+            this.context.allow = value !== false;
```
Sumragen: Isn't looks like "Doc improvement" :)

PR #333, opened 2020-04-13
Fix/unresolved subscription

Sumragen:


src/app/shared/components/xm-password-needed/xm-password-needed.component.ts
```diff
@@ -27,7 +28,7 @@ export class XmPasswordNeededComponent implements OnInit {
     }
 
     public ngOnInit(): void {
-        this.eventManager.subscribe('error.passwordNeeded', (event) => {
+        this.eventManagerSubscription = this.eventManager.subscribe('error.passwordNeeded', (event) => {
```
didyk, line 31: Why do you not use takeUntilOnDestroy()?

Sumragen: Because it isn't classical Observable.
Check method `subscribe` of `EventManagerService`. This method returns `Subscription` instead of `Observable`.
Hope it's an answer for your question.

PR #352, opened 2020-04-22
Develop sidebar right

Kaskyi:


src/app/modules/xm-sidebar-right/sidebar-right.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    public setContainer(container: Container): void {
+        this.container = container;
+    }
+
+    public removeContainer(): void {
+        this.container = null;
+    }
+
+    public open<T = unknown, D = unknown>(
+        componentOrTemplateRef: ComponentType<T> | TemplateRef<T>,
```
didyk, line 25: We have types and we don't need to describe it in the variable name

src/app/modules/xm-sidebar-right/xm-sidebar-right.module.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    public create<T, D>(templateRef: TemplateRef<D> | Type<T>, config: SidebarRightConfig<D>): T | null {
+        const viewContainerRef = this.xmContainerOutlet.viewContainerRef;
+        if (viewContainerRef.length > 0) {
+            viewContainerRef.clear();
+        }
+
+        if (templateRef instanceof TemplateRef) {
+            viewContainerRef.createEmbeddedView(templateRef);
+            this.openStyles(config.width || '300px');
```
didyk, line 55: Could we move '300px' to constants?

Kaskyi: done

src/app/modules/xm-sidebar-right/sidebar-right.service.ts
Kaskyi: вщту

PR #360, opened 2020-04-27
chore(xm-permission): add empty value

Kaskyi:
#357


src/app/shared/privilege/xm-permission.service.ts
```diff
@@ -29,7 +29,7 @@ export class XmPermissionService {
 
     public get permissions$(): Observable<XmUserPermission[]> {
         return this.userService.user$.pipe(
-            filter(Boolean),
+            map(u => u || {permissions: []}),
```
Sumragen, line 32: Check RxJS operator `defaultifempty`. Probably, it's much clean approach.

Kaskyi: done

PR #386, opened 2020-05-06
Feature/#357/server url

Kaskyi:
#357


Kaskyi:
add interceptor instead

PR #493, opened 2020-06-15
feat(navbar): added flag to uiConfig for toggling full match search in header

ozelenyi:


src/app/layouts/navbar/navbar.component.ts
```diff
@@ -125,7 +127,8 @@ export class NavbarComponent implements OnInit, DoCheck {
 
     public search(term: string): void {
         if (term) {
-            this.router.navigate(['/search'], {queryParams: {query: term, dashboardId: this.getDashboardId()}});
+            const searchQuery = this.searchFullMatch ? `"${term}"` : term;
```
andrewkikot, line 130: `"${term}"` shouldn't this be with single quotes

ozelenyi: No, it`s double for a reason

> `"${term}"` shouldn't this be with single quotes?

No, it`s double for a reason

PR #654, opened 2020-08-12
fix(mat-dialog): adding controls

andrewkikot:
this PR adds controls to mat-dialog


src/app/xm-entity/entity-detail-dialog/entity-detail-dialog.component.html
```diff
@@ -4,6 +4,9 @@
                 hasHeaderConfig else
                 hasNotHeaderConfig;">
     </div>
+    <button class="mat-dialog-close" type="button" tabindex="-1" mat-icon-button (click)="onCancel()">
```
andrewkikot, line 7: move to separate component

PR #656, opened 2020-08-12
fix(linked-in): fix translates key

Kaskyi:


ishkurko:
Propose to remove linkedin-profile from core.

PR #663, opened 2020-08-13
feat(string-handler): add touch event

Kaskyi:


packages/components/src/string-handler/string-handler.directive.ts
```diff
@@ -28,8 +28,21 @@ export class StringHandlerDirective extends NgModelWrapper<string> {
     @HostListener('focusout')
     public applyHandlers(): void {
         const value = this.elementRef.nativeElement.value;
-        if (!value) return;
+        if (value) {
         this.elementRef.nativeElement.value = this.proceed(value);
+        }
         this.change(this.elementRef.nativeElement.value);
```
Sumragen, line 34: We will apply function `change` with parameter of `nativeElement.value` previous value.
Is that correct?

Kaskyi: Ive fixed it with the writeValue method

PR #725, opened 2020-09-03
fix(button-groups): fixing inline spacing, refactoring font sizing

andrewkikot:


packages/styles/src/_button-group.scss
```diff
@@ -5,14 +5,17 @@
 
 @mixin xm-button-group($theme) {
   .xm-button-group {
+    font-size: 0;
```
Kaskyi, line 8: why do u set 0 ?

Kaskyi (approved):

packages/styles/src/_button-group.scss
andrewkikot: this is one of techniques for removing inline-block html spacing

brigart (approved):

PR #742, opened 2020-09-28
fix(b2bcrm787): fix if component dont have registerOnChange

EKrasovskiy:


packages/xm-dynamic/src/control/dynamic-control.directive.ts
```diff
@@ -140,7 +141,7 @@ export class DynamicControlDirective<V, O>
     }
 
     protected updateRegisterOnChange(): void {
-        if (!this.instance) {
+        if (!this.instance || typeof this.instance.registerOnChange !== 'function') {
```
Sumragen, line 144: Probably there will be an error in case when first part of predicate is true.
if "this.instance" isn't defined then "this.instance.registerOnChange" will produce "Cannot read property 'registerOnChange' of undefined"

Sumragen (changes requested):

packages/xm-dynamic/src/control/dynamic-control.directive.ts
EKrasovskiy: No, it can`t produse an error. If first part of (||) operator will be true, second part will never be execute.

Sumragen: Agree

Sumragen (approved):

PR #747, opened 2020-09-29
fix(b2bcrm-800) add filter by unique id and hide TypeError

EKrasovskiy:


Kaskyi (approved):

packages/administration/src/admin.service.ts
```diff
@@ -100,10 +101,11 @@ export class BaseAdminListComponent implements OnInit, OnDestroy {
     }
 
     public onSuccess(data: any, headers: any): any {
```
Sumragen, line 103: We could specify interface of `data` parameter. At least we knew property `id`

Sumragen (changes requested):

packages/administration/src/admin.service.ts
EKrasovskiy: yes, ok.

Sumragen:
Also specify all typings that we know, please.
image

Sumragen (approved):

PR #802, opened 2020-10-30
feat(25529): added logic to handle "readOnlyConfig" flag in UI specification

brigart:
When flag is TRUE actions (except filters, sort, pagination, etc.) will be blocked for next components:

- specification-mng
- roles-matrix
- roles-management
- roles-management-detail


brigart:
I'll add flag description into wiki after merge soon

andrewkikot (approved):

Kaskyi (approved):

PR #838, opened 2020-11-20
chore(date): update date naming

Kaskyi:


packages/components/date/examples/xm-date-example.ts
```diff
@@ -0,0 +1,43 @@
...
+        <xm-date-control [value]="dateValue" [options]="dateControlOptions"></xm-date-control>
+        <xm-date-range-control [value]="dateRangeValue" [options]="dateViewOptions"></xm-date-range-control>
+    `,
+})
+export class XmDateExample {
+    public dateValue: XmDateValue = new Date();
+    public dateOptions: XmDateOptions = { format: 'dd.MM.yyyy HH:mm' };
+    public dateRangeValue: XmDateRangeControlValue = { from: new Date(), to: new Date() };
+    public dateViewOptions: XmDateViewOptions = { title: 'date-view', ...this.dateOptions };
+    public dateControlOptions: XmDateControlOptions = { title: 'date-control-view', name: 'date-control-view' };
```
Sumragen, line 28: It looks like word date is redundant in property name due to we have described domain in component name.

packages/components/date/testing/xm-date-testing.module.ts
```diff
@@ -0,0 +1,37 @@
...
+})
+export class MockXmDateView {
+}
+
+
+@NgModule({
+    exports: [MockXmDateComponent, MockXmDateControl, MockXmDateRangeControl, MockXmDateView],
+    declarations: [MockXmDateComponent, MockXmDateControl, MockXmDateRangeControl, MockXmDateView],
+})
+export class XmDateTestingModule {
```
Sumragen, line 36: Sounds like we will import some part of logic that will manipulate/calculate date objects

packages/components/date/xm-date-view.ts
```diff
@@ -0,0 +1,43 @@
...
+            <span xmLabel>{{options?.title | translate}}</span>
+            <xm-date xmValue [value]="value" [options]="options"></xm-date>
+        </xm-text>
+    `,
+    changeDetection: ChangeDetectionStrategy.Default,
+})
+export class XmDateView implements IComponent<XmDateValue, XmDateViewOptions>, OnInit, OnChanges {
+    @Input() public value: XmDateValue;
+    @Input() public options: XmDateViewOptions;
+    public styleInline: boolean;
```
Sumragen, line 25: could we remove that property? It looks like we could retrieve necessary value from config directrly.

packages/components/date/date.md
```diff
@@ -0,0 +1,4 @@
+`xm-date` - date value
+`xm-date-view` = date view
```
Sumragen, line 2: should symbol `=` to be `-`?

Sumragen (changes requested):

packages/components/date/date.md
Kaskyi: +

packages/components/date/xm-date-view.ts
Kaskyi: no - where will be a typing error

packages/components/date/examples/xm-date-example.ts
Kaskyi: `value` and `options` won't be self-described and it is easier to tell which component it belongs to

packages/components/date/testing/xm-date-testing.module.ts
Kaskyi: \+

packages/components/date/examples/xm-date-example.ts
Sumragen: why is it? In case when I use component `xm-date-example` it quite clearly for me that value of it will be `Date`, isn't? As an example when we use component `username-label` it's quite clear that value is some username.

packages/components/date/xm-date-view.ts
Sumragen: which one? We could use `!!` syntax to transform value.

Sumragen (changes requested):

EKrasovskiy (approved):

Sumragen (approved):

PR #962, opened 2021-03-05
Support url relations

sergeysenja1992:


src/app/shared/jsf-extention/widgets/ext-select/ext-select.component.ts
```diff
@@ -244,6 +193,22 @@ export class ExtSelectComponent implements OnInit, OnDestroy, AfterViewInit {
         }
     }
 
+    private filterElements(): void {
+        if (!this.elements) {
+            return;
+        }
+        let search = this.elementFilterCtrl.value;
+        if (!search && search == null) {
```
brigart, line 201: `search == null` will be true if search is undefined, is it ok here?

sergeysenja1992: it's some magic, i just move this function without editing =))

brigart (approved):

andrewkikot (approved):

PR #985, opened 2021-04-05
feat(logger): add xm logger service

Kaskyi:


packages/logger/src/xm-logger.service.ts
```diff
@@ -0,0 +1,73 @@
+import { Injectable } from '@angular/core';
```
oscaryald, line 1: maybe should be better to separate classes into files

oscaryald (approved):

packages/logger/src/xm-logger.service.ts
Kaskyi: done

oscaryald (approved):

EKrasovskiy (approved):

Sumragen (approved):

PR #1019, opened 2021-04-27
fix(table-column-dynamic-cell): clear border-bottom on cdk-focused

Vlad-Gordienko:


packages/components/table/column/xm-table-column-dynamic-cells.ts
```diff
@@ -77,6 +77,12 @@ export const XM_TABLE_COLUMN_DYNAMIC_CELLS_OPTIONS_DEFAULT: XmTableColumnDynamic
             </td>
         </ng-container>
     `,
+    styles: [`
+        .mat-sort-header-container {
```
Kaskyi, line 81: the style changes should be local in your theme or component

Kaskyi (changes requested):

PR #1023, opened 2021-05-07
feat(enum-control): show default value in selection if property empty

EKrasovskiy:


EKrasovskiy:
to use this feature you should add in options.items element without value

grankovsky (approved):

Kaskyi (approved):

PR #1057, opened 2021-05-27
fix(format): correct array use in format

Vlad-Gordienko:


Kaskyi (approved):

packages/shared/operators/src/format.ts
```diff
@@ -40,17 +40,25 @@ import * as _ from 'lodash';
...
-export function format<T>(template: object | unknown, entity: unknown): T {
-    const res = {} as T;
-    for (const key in template as object) {
-        if (Object.prototype.hasOwnProperty.call(template, key)) {
-            const value = template[key];
-            if (typeof value === 'string') {
-                res[key] = interpolate(value, { entity, _ });
-            } else if (typeof value === 'object') {
-                res[key] = format(value, entity);
+export function format<T extends Array<any>>(template: object | unknown, entity: unknown): T {
```
Sumragen, line 43: T extends Array<any> is redundant there. We could specify return type like T | T[]

packages/shared/operators/src/format.ts
```diff
@@ -40,17 +40,25 @@ import * as _ from 'lodash';
...
+    if (template instanceof Array) {
+        const res = [] as T;
```
Sumragen, line 45: as T[]

Sumragen (changes requested):

Sumragen (approved):

PR #1088, opened 2021-06-16
Fix error prop from null

andrewkikot:


Kaskyi (approved):

Sumragen (dismissed):

andrewkikot:
Since [this commit](https://github.com/xm-online/xm-webapp/commit/bf38e40f30824be9fbc85f903967522f9ba2a15f) there are changes, which breaking idp. That is happening because in int() of auth-jwt.service.ts its now clear all stored configs from session storage, and we need some of them to handle idp redirect.

Kaskyi (approved):

Sumragen (approved):

PR #1120, opened 2021-07-06
fix(permission): replace takeUntil with subscription

Kaskyi:


EKrasovskiy (approved):

packages/core/permission/src/permission.directive.ts
```diff
@@ -104,12 +104,11 @@ export class PermissionDirective implements OnInit, OnDestroy {
     }
 
     public ngOnDestroy(): void {
-        takeUntilOnDestroyDestroy(this);
+      this.subscription.unsubscribe();
     }
 
     public ngOnInit(): void {
-        this.update$.pipe(
-            takeUntilOnDestroy(this),
```
Sumragen, line 112: is that a reason why we need to remove `takeUntilOnDestroy` at all?

Sumragen (changes requested):

packages/core/permission/src/permission.directive.ts
Kaskyi: takeUntilOnDestroy in this case produces a memory leak

Sumragen (approved):

PR #1152, opened 2021-07-22
refactor(core-theme): update structure

Kaskyi:


EKrasovskiy (approved):

Sumragen (approved):

packages/core/theme/src/services/xm-dark-theme-manager.ts
```diff
@@ -2,7 +2,7 @@ import { Injectable, OnDestroy } from '@angular/core';
 import { Subscription } from 'rxjs';
 import { Observable } from 'rxjs';
 import { ThemeSchemeService } from './theme-scheme.service';
-import { THEME_STRATEGY, XmTheme, XmThemeStore } from './xm-theme-store.service';
+import { THEME_STRATEGY, XmTheme, XmThemeStore } from '../stores/xm-theme-store.service';
```
sergeysenja1992, line 5: how about backward compatibility with existing extensions that extend or use this theme?

Kaskyi: All imports should use '@xm-ngx/core/theme' path

grankovsky (approved):

PR #1176, opened 2021-07-29
fix(web-app): remove incorrect value change

grankovsky:


Kaskyi (approved):

EKrasovskiy (approved):

src/app/shared/privilege/xm-permitted.directive.ts
```diff
@@ -43,7 +43,6 @@ export class XmPermittedDirective implements OnInit, OnDestroy, AfterContentInit
     public set xmPermitted(value: string[]) {
         this._xmPermitted = typeof value === 'string' ? [value] : value;
```
Sumragen, line 44: Are we really sure `value` has type `string`? It looks like we need to define correct function signature...

Kaskyi: its backward compatibility, unfortunately, developers sending string values and creates a problem.

avchugaev: According to logic, `value`'s type must be `string | string[]`

Kaskyi: this discussion isn';t a part of this PR.  Please create an issue and it will be resolved soon

avchugaev (approved):

PR #1200, opened 2021-08-15
fix(attachment-detail-dialog) fix can not read property attachment.co…

Vlad-Gordienko:
…ntent.value on attach


Kaskyi (approved):

EKrasovskiy (approved):

src/app/xm-entity/attachment-detail-dialog/attachment-detail-dialog.component.ts
```diff
@@ -46,14 +46,13 @@ export class AttachmentDetailDialogComponent implements OnInit {
     }
 
     public ngOnInit(): void {
-        if (!this.attachmentSpecs || !this.attachmentSpecs.length) {
-            return;
+        if (this.attachmentSpecs[0]) {
```
sergeysenja1992, line 49: так а если бек вернул поле attachmentSpecs в значение null

Kaskyi: there are two cases:
1. An interface declares only string arrays `AttachmentSpec[]`
2. Reality is - it is possible to get `null` or `undefined`

pls add null and undefined to field type.
add data validation at ngOnInit

grankovsky (approved):

PR #1222, opened 2021-09-09
fix(xm-control-errors-translates): add default errors translates

oscaryald:


Kaskyi (commented):
valueLessThanIn is not the context of  xm-control-errors-translates.ts

PR #1245, opened 2021-10-04
feat(xm-text-control): add possibility to trim value

oscaryald:


Vlad-Gordienko (commented):
create a new control and extends it from text-control

PR #1264, opened 2021-11-04
fix: returning browser lang if only included to supported langs

andrewkikot:


src/app/modules/xm-translation/getBrowserLocale.ts
```diff
@@ -19,5 +21,5 @@ export function getBrowserLocale(): string | null {
     if (browserLang.indexOf('_') !== -1) {
         browserLang = browserLang.split('_')[0];
     }
-    return browserLang;
```
Kaskyi, line 22: the method should returns the browse locales not the customs locales

Kaskyi: somewhere must be a service that checks for supported languages

andrewkikot: ok, will think about right place for this logic

staciedev (approved):

PR #1340, opened 2022-04-22
feat(date-range): make date range work as dynamic widget

zverbeta:


packages/components/date/xm-date-range-control.ts
```diff
@@ -30,47 +59,180 @@ export interface XmDateRangeControlValue {
...
+    ) {
+        super(ngControl);
+    }
+
+    private update(): void {
+        if (this.value && this.options) {
+            this._value = this.toModel(this.value);
+        }
+
+        if (this.value == null) {
```
oscaryald, line 126: better use strict "===" or in this case can simplify if(!this.value)

Sumragen (dismissed):

packages/components/date/xm-date-range-control.ts
```diff
@@ -30,47 +59,180 @@ export interface XmDateRangeControlValue {
...
+        ).subscribe((dates) => {
+            const value = this.fromModel(dates);
+
+            this.change(value);
+        });
+    }
+
+    private toModel(value: XmDateRangeValueOrString): XmDateRangeControlValue {
+        if (typeof value === 'string') {
+            if (this.options?.transform == null) {
```
oscaryald, line 151: better use strict "===" or in this case can simplify if(!this.options?.transform)

packages/components/date/xm-date-range-control.ts
```diff
@@ -30,47 +59,180 @@ export interface XmDateRangeControlValue {
...
+                    from: null,
+                    to: null,
+                };
+            }
+
+            const [from, to] = value
+                ?.split(` ${this.getSeparator()} `)
+                ?.map(x => _.trim(x, this.buildQuotedString(' ')));
+
+            if (from == null && to == null) {
```
oscaryald, line 162: same

packages/components/date/xm-date-range-control.ts
```diff
@@ -30,47 +59,180 @@ export interface XmDateRangeControlValue {
...
+            }
+
+            return this.normalizeDates({ from, to });
+        }
+
+        return value;
+    }
+
+    private fromModel(dates: XmDateRangeControlValue): string {
+        if (this.options?.transform == null) {
```
oscaryald, line 176: same

oscaryald (approved):

packages/components/date/xm-date-range-control.ts
zverbeta: strict comparison compare only null check, but i want to check undefined or null

`let options;

options?.transform === null // false
options?.transform == null`

oscaryald: ok, than can simplify if(!this.value)

zverbeta: +

Sumragen (approved):

zverbeta:
@oscaryald wait for your approval

Kaskyi (approved):

PR #1350, opened 2022-04-29
feat(support-http-total): support headers total count

zverbeta:


packages/components/entity-collection/http-client.rest.ts
```diff
@@ -71,7 +71,7 @@ export class HttpClientRest<T extends IId = unknown, Extra extends Pageable = Pa
         const extra = {
             pageIndex: 0,
             pageSize: res.body?.length || 0,
-            total: res.body?.length || 0,
```
Kaskyi, line 74: use XmEntityRepository instead

Kaskyi (changes requested):

grankovsky (approved):

PR #1357, opened 2022-05-12
feat(presentation-options-and-config): render dynamic widget with config

zverbeta:
Render dynamic widget config with input name `config`

### DIFF
--- a/packages/dynamic/src/presentation/xm-dynamic-presentation-base.directive.ts
+++ b/packages/dynamic/src/presentation/xm-dynamic-presentation-base.directive.ts
@@ -18,10 +18,22 @@ interface IValue<V> {
     value: V;
 }
 
-/** Determines input(control) options. */
+/**
+ * @deprecated
+ * Determines input(control) options.
+ **/
 interface IOptions<O> {
-    /** Input options. */
-    options: O;
+    /**
+     * @deprecated
+     * Input options.
+     **/
+    options?: O;
+}
+
+/** Determines input(control) config. */
+interface XmDynamicConfig<O> {
+    /** Input config. */
+    config?: O;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


Sumragen (changes requested):
Shall `PresentationBase` be used as parent class for widgets?

zverbeta:
PresentationBase used to not only for widgets, but controls to. Dynamic widgets can accept both `config` and `options`

packages/dynamic/src/presentation/xm-dynamic-presentation-base.directive.ts
```diff
@@ -24,16 +24,24 @@ interface IOptions<O> {
     options: O;
 }
 
+/** Determines input(control) config. */
+interface IConfig<O> {
+    /** Input config. */
+    config?: O;
```
Kaskyi, line 30: Do u want to render widgets inside presentation components ?

Kaskyi (changes requested):

packages/dynamic/src/presentation/xm-dynamic-presentation-base.directive.ts
zverbeta: Yes, because presentation directive render all types of dynamic components: widgets, components, controls

Kaskyi: presentation/controls - is a part of dynamic feature
widget/dashboard is a part of dashboard feature
dynamic-widget is combination of dashboard and dynamic features.
config it is a property of widget not a presentation. We should add one more abstract layer without configs and options.

Kaskyi (approved):

major13ua (approved):

sergeysenja1992 (approved):

Sumragen (approved):

Sumragen:
Will discuss later which field shall we left

PR #1372, opened 2022-05-30
fix(xm-text-control): mat-hint styles fixed

LevBD:


packages/components/text/text-control/xm-text-control.ts
```diff
@@ -72,6 +72,9 @@ const XM_TEXT_CONTROL_OPTIONS_DEFAULT: XmTextControlOptions = {
 
         </mat-form-field>
     `,
+    styles: [
+        '.mat-hint { max-width: fit-content }',
```
Kaskyi, line 76: ViewEncapsulation.None - all styles become global

Kaskyi (changes requested):

packages/components/text/text-control/xm-text-control.ts
LevBD: Changed styles to inline styles

Kaskyi (dismissed):

Sumragen (dismissed):

Sumragen (dismissed):

Kaskyi (approved):

PR #1402, opened 2022-06-30
Feat/calendar from master branch

opishchela:


packages/styles/src/_scroll.scss
```diff
@@ -31,3 +31,8 @@ $xm-scrollbar-background-color: #cbcbcb !default;
     scrollbar-track-color: $xm-scrollbar-background-color;
   }
 }
+
+.cdk-global-overlay-wrapper {
```
Kaskyi, line 35: move to https://github.com/xm-online/xm-webapp/blob/main/packages/styles/src/vendor/_angular-material.scss

sergeysenja1992 (approved):

PR #1430, opened 2022-07-28
feat(http-client-patch): add http client patch method

zverbeta:


grankovsky (dismissed):

packages/components/entity-collection/i-entity-collection-pageable.ts
```diff
@@ -43,6 +43,9 @@ export interface IEntityCollectionPageable<T extends IId = unknown, Extra extend
     /** PUT request. Use to replace or delete entity data. */
     update(update: Partial<T>, params?: QueryParams, headers?: HttpHeaders): Observable<HttpResponse<T>>;
 
+    /** PATCH request. Use to replace or delete entity data. */
```
Kaskyi, line 46: Why do we need a new method if the `update` method do the same thing ?

Bozonych: When we want update one field of entity:
`update` -  we must send all entity
`path` - we can send only updated field

Bozonych (dismissed):

packages/components/entity-collection/i-entity-collection-pageable.ts
Kaskyi: There are 2 points:
- tsdoc comment are same for update and patch.
- update and patch argument is Partial<T>.
In http there put and patch. In collection I suggest to change to `update(patch)` and `replace(put)`

sergeysenja1992 (dismissed):

packages/components/entity-collection/i-entity-collection-pageable.ts
sergeysenja1992: PUT and PATCH is different http method, PATCH very usefull. We plan to support it on entity BE too.

Sumragen (dismissed):

PR #1431, opened 2022-08-02
fix(xm-text-range-control): fixed mat-hint styles

LevBD:
Screenshot from 2022-08-02 08-37-01

### DIFF
--- a/packages/components/text/text-range-control/xm-text-range-control.component.ts
+++ b/packages/components/text/text-range-control/xm-text-range-control.component.ts
@@ -54,7 +54,7 @@ const XM_TEXT_RANGE_CONTROL_OPTIONS_DEFAULT: XmTextRangeControlOptions = {
 
             <mat-error *xmControlErrors="control?.errors; message as message">{{message}}</mat-error>
 
-            <mat-hint *ngIf="options.maxLength" align="end">{{value?.length}} / {{options.maxLength}}</mat-hint>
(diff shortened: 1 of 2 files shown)


packages/components/text/text-range-control/xm-text-range-control.component.ts
```diff
@@ -54,7 +54,7 @@ const XM_TEXT_RANGE_CONTROL_OPTIONS_DEFAULT: XmTextRangeControlOptions = {
 
             <mat-error *xmControlErrors="control?.errors; message as message">{{message}}</mat-error>
 
-            <mat-hint *ngIf="options.maxLength" align="end">{{value?.length}} / {{options.maxLength}}</mat-hint>
+            <mat-hint *ngIf="options.maxLength" align="end" style="min-width: fit-content">{{value?.length}} / {{options.maxLength}}</mat-hint>
```
oscaryald, line 57: might be better remove style to scss file

LevBD: I didn't quite understand what you mean. This component does not have a separate file with styles.

oscaryald: I mean don't use inline style in html, not good practice. better move it to separate scss file or add to @Component decorator. Ex:     style: [
        `.mat-hint {
           min-width: fit-content
        }`   
    ]

Sumragen (approved):

oscaryald (approved):

PR #1444, opened 2022-09-23
feat(multi-select): multi select control

zverbeta:


Sumragen (approved):

packages/components/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,237 @@
...
+    valueKey: string;
+    titleKey: string;
+}
+
+type XmMultiSelectItemOrString = XmMultiSelectItem | string;
+
+@Component({
+    selector: 'xm-multi-select-control',
+    template: `
+        <mat-form-field [ngClass]="{ 'mat-form-field-disabled': disabled }" class="label-touchable">
```
sergeysenja1992, line 43: maybe better to move to separate file

zverbeta: If you take popular component libraries, for example (material), you can see their single file components approach. Also in the new version of Angular (14) you can see the same approach, when the components have become autonomous (standalone).

P / S If you break the components into smaller ones, then this approach looks more successful

Kaskyi (approved):

grankovsky (approved):

andrewkikot (approved):

PR #1460, opened 2022-10-21
fix(fix-selected-states): fix selected states

zverbeta:


packages/administration/src/specification-management/entity-spec-editor/states-editor/states-editor.component.ts
```diff
@@ -33,19 +33,18 @@ export class StatesEditorComponent extends EditorUtils {
...
-            });
-        } else {
-            stateSpec.next = stateSpec.next.filter((n) => n.stateKey !== event.option.value.key);
+        if (!stateSpec.next) {
+            stateSpec.next = [];
         }
-    }
+        const selectedStateSpec = event.options.map((option) => ({
+            stateKey: option.value.key,
+            name: JSON.parse(JSON.stringify(option.value.name)),
```
Bozonych, line 42: Maybe can, don`t clone name ? or use {...name}

packages/administration/src/specification-management/entity-spec-editor/states-editor/states-editor.component.ts
```diff
@@ -33,19 +33,18 @@ export class StatesEditorComponent extends EditorUtils {
...
+        }));
+
+        const mergeStateSpec = [...stateSpec.next, ...selectedStateSpec];
+        const excludeSelectedStateSpec = stateSpec.next.filter((s1) => selectedStateSpec.some(s2 => s1.stateKey === s2.stateKey));
 
+        stateSpec.next = mergeStateSpec.filter(m => !excludeSelectedStateSpec.some(e => m.stateKey === e.stateKey));
```
Bozonych, line 48: it's sou complicated code. 
Maybe we can simplify it? Like that`s:
`
const selectedStateKeys = selectedStateSpec.map(state => state.stateKey);
stateSpec.next = stateSpec.next.filter(state =>  !selectedStateKeys.includes(state.stateKey))`
or another solution but without using Array.some function.

what is your opinion about this?

zverbeta: this method rarely used, so using "Array.some" function possible here

packages/administration/src/specification-management/entity-spec-editor/states-editor/states-editor.component.ts
zverbeta: Legacy. I agree this convert looks weird.

Bozonych (approved):

grankovsky (approved):

PR #1497, opened 2022-12-27
feat(jsf): add different library

Kaskyi:
Signed-off-by: Yevhenii Kamenskyi <yevhenii.kamenskyi@gmail.com>


sergeysenja1992 (approved):

packages/json-schema-form/core/xm-json-schema-form.module.ts
```diff
@@ -97,7 +102,7 @@ export const ANGULAR_MATERIAL_MODULES = [
         },
     ],
 })
-export class XmJsonSchemaFormModule {
+export class OldXmJsonSchemaFormModule {
```
sergeysenja1992, line 105: pls do not rename module that using in existing projects, better to mark as deprecated and give other name for new module.

PR #1500, opened 2022-12-28
feat(class-style-literals): class & style template

zverbeta:


packages/dynamic/src/cell/xm-dynamic-cell.directive.ts
```diff
@@ -82,6 +83,26 @@ export class XmDynamicCellDirective<V, O extends XmDynamicCell<O>>
         return getValue(this.row, this._cell.field);
     }
 
+    private getStyle(style?: string): string {
+        if (style?.includes('${')) {
+            try {
+                style = _.template(style ?? '')(this.row as object ?? {});
+            } catch(e) {}
```
Kaskyi, line 90: 1. Is there a way to prevent getting an invalid template?
2. Please provide a solution to notify developer or configurator about invalid string.

zverbeta: 1. No
2. console.warn will fit?

Kaskyi: okey

Kaskyi (approved):

Bozonych (approved):

PR #1555, opened 2023-02-06
refactor(cli): remove ext- prefix

Kaskyi:
BREAKING_CHANGES
all selectors starts with `ext-*` should be renamed to module name

Signed-off-by: Yevhenii Kamenskyi <yevhenii.kamenskyi@gmail.com>


sergeysenja1992 (changes requested):
backward compatibility at the configuration level must be preserved in any situation, we cannot allow dashboards to stop working when updating the ui

sergeysenja1992 (changes requested):
this is unacceptable change

PR #1639, opened 2023-04-14
feat(confirm-dialog): fix confirm dialog empty controls

zverbeta:


Bozonych (approved):

Kaskyi (approved):

packages/dynamic/control/xm-dynamic-control.directive.ts
```diff
@@ -83,7 +88,16 @@ export class XmDynamicControlDirective<V, O>
     
 
     public ngOnInit(): void {
-        this.createComponent().then();
+        from(this.createComponent()).pipe(
```
Kaskyi, line 91: why do you pipe from promise and switchMap it ? You can just this.controlValue.asObservable().pipe(

Kaskyi (changes requested):

packages/dynamic/control/xm-dynamic-control.directive.ts
zverbeta: Because we needs a control instance

Kaskyi: when user change selector / component recreates https://github.com/xm-online/xm-webapp/pull/1639/files#diff-6ddc14443726705185dd92c39c15da1745c9585ae0f51d3dd06044dcb9e2ae22R117

Kaskyi (approved):

grankovsky (approved):

PR #1671, opened 2023-05-17
Refactor/core dependencies

Kaskyi:
Move login service.
```ts
import { LoginService } from '@xm-ngx/core/auth';
import { LoginService } from '@xm-ngx/components/login';
```


packages/core/theme/src/stores/xm-theme-store.service.ts
```diff
@@ -12,6 +12,12 @@ import { XmTheme } from '../interfaces/xm.theme';
  */
 export class XmThemeStore {
     public get(): XmTheme | null {
+
+        // TODO:TEMPORARY
+        //  While user theme store not supported
+        //  the main way to set a theme is provide it via ui config
+        return null;
```
sergeysenja1992, line 19: what reason to do it?

packages/core/user/src/account.service.ts
```diff
@@ -45,11 +44,11 @@ export class AccountService {
         return this.http.post(SERVER_API_URL + ACCOUNT_TFA_DISABLE_URL, {}, {observe: 'response'});
     }
 
-    public getProfile(req?: any): Observable<XmEntity> {
+    public getProfile(req?: any): Observable<any> {
```
sergeysenja1992, line 47: this method must be typed, why we make code worse?

packages/entity/src/entity-list-card/entity-list-card-options.model.ts
```diff
@@ -64,8 +47,8 @@ export interface ActionsOptions extends ActionOptions {
     navigateByInnerUrl?: string;
     functionKey?: string;
     privilege?: string[] | string;
-    handler?: (entity: XmEntity) => any;
-    actionCondition?: (entity: XmEntity) => any;
+    handler?: (entity: IId) => any;
```
sergeysenja1992, line 50: this is not equals replacement

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "xm-webapp",
-  "version": "4.0.64",
+  "version": "4.0.65",
```
sergeysenja1992, line 3: pls increment second number
it's not a minor changes, it`s breaking changes

packages/core/user/src/account.service.ts
Kaskyi: first i want to split /core package with other packages, complete packages build, than refactor and provide property dependencies between packages.

package.json
Kaskyi: done

PR #1712, opened 2023-06-28
chore(fixes): some fixes

zverbeta:


Kaskyi (approved):

denisogolubev:
According to the Code style guide templates and styles that have more than 3 lines of code should be in their own files.
https://angular.io/guide/styleguide#extract-templates-and-styles-to-their-own-files

So, I think it's not necessary to merge most of these changes.

grankovsky (approved):

zverbeta:
@denisogolubev  

Usually, the rules described in the documentation for any library or framework are just a recommendation, rather than a real example of how to do it.

While it is generally recommended to maintain clean and modular code, the decision to extract templates and styles into separate files should depend on the specific context and the complexity of the code.

denisogolubev:
@zverbeta it's true that it's just a recommendation.
But I think this argument is strong enough to follow this rule:
`Large, inline templates and styles obscure the component's purpose and implementation, reducing readability and maintainability.`

zverbeta:
@denisogolubev 

I don't think so, here are some reasons why.

1. The documentation is written for thousands of users of various levels, so the documentation provides basic information
2. The documentation has a recommendation on how to do it better, but the company has certain guidelines for writing and maintaining the code.

In situations where there are disputes or conflicting recommendations, it is crucial to evaluate the context and requirements of the project at hand.

denisogolubev:
@zverbeta it's true, you can modify the rules for the team's needs.
All files should be responsible for the parts of your application that they were created for.
For example, you can create a `service` for the business logic, but if look at this PR all can be placed in one file.
How do you think it's better to support code when you have 5-6 files with 50-100 lines of code for each of them or one file where will be 250-600 rows of code?

Angular is all about project structuring and pieces that are responsible for certain business/application logic.
https://angular.io/guide/styleguide#file-structure-conventions

Even if you run `ng g c some-component` you will get 4 files.
Angular `kindly` ask you to follow this rules. Because when someone wants to work with you in the future it will be easier to understand your project and you will spend less time explaining to people what is happening in your code. Just because you will follow the framework recommendations.

zverbeta:
@denisogolubev 

In this case, I don't see the code, which would be at least approximately 600 lines of code, can you please show me?

> Even if you run ng g c some-component you will get 4 files.

Otherwise we can configure it in angular.json or via command line options via --inline-style, --inline-template

As I said, the documentation is only a recommendation, I think that you should be guided by a specific case


I would like to hear other comments if there are any

denisogolubev:
@zverbeta I just want to let you know that this approach is not about an expandable code base.
Today your file has 100 rows of code. In a month someone will add a new functionality and it will be 150 rows of code and so on.
So, in the future, you might need to separate these components into different files, which is why I'm sure that this pattern is not about flexibility and expandability.
And looks like it's not necessary refactoring.

Sumragen (approved):

PR #1722, opened 2023-07-06
feat(alert-dialog): drop sweetalert

zverbeta:
We want to remove sweetalert from dependencies and migrate to modal dialogs from material design.

In this implementation, we support the following properties: https://github.com/xm-online/xm-webapp/blob/feat/alert-dialog/packages/alert/src/xm-alert.interface.ts#L3

If your project uses the `input` property from sweetalert. 

You need to use this service instead https://github.com/xm-online/xm-webapp/blob/main/packages/components/confirm-dialog/src/confirm-dialog.service.ts#L21


Kaskyi (approved):

Kaskyi (approved):

grankovsky (approved):

packages/alert/src/xm-alert.component.scss
```diff
@@ -0,0 +1,45 @@
...
+    display: flex;
+    flex-direction: column;
+
+    &.center {
+      @include center();
+    }
+
+    &-icon {
+      line-height: 1;
+      margin-bottom: 16px;
```
Bozonych, line 17: Do not use px for typography, make sense use bootstrap class like: mb-2.

Bozonych (approved):

packages/alert/src/xm-alert.component.scss
```diff
@@ -0,0 +1,45 @@
+@mixin center {
```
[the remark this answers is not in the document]
zverbeta: In latest bootstrap version we have text-center utility class. but i want to use 1 class for center childs (not just text)

packages/alert/src/xm-alert.component.scss
zverbeta: I see on the other hand we would like to customize it later according to material specifications, so using a bootstrap class would be inappropriate

Sumragen (approved):

PR #1728, opened 2023-07-16
Get rid of widget 'options'

Sumragen:
Use `config` instead of `options`.
Added handler for default value of BaseAdminListComponent

### DIFF
--- a/packages/administration/src/admin.service.ts
+++ b/packages/administration/src/admin.service.ts
@@ -18,7 +18,7 @@ import { NotImplementedException } from '@xm-ngx/shared/exceptions';
 @Directive()
 export class BaseAdminListComponent implements OnInit, OnDestroy {
 
-    public config: {
+    private actualConfig: {
         pageSizeOptions: number[],
         pageSize: number,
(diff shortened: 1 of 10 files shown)


Bozonych (approved):

packages/administration/src/admin.service.ts
```diff
@@ -17,19 +18,38 @@ import { NotImplementedException } from '@xm-ngx/shared/exceptions';
...
 export class BaseAdminListComponent implements OnInit, OnDestroy {
 
-    public options: {
+    private actualConfig: {
         pageSizeOptions: number[],
         pageSize: number,
         sortDirection: 'asc' | 'desc',
         sortBy: string,
         navigateUrl: string[];
+        columns?: XmTableColumn[]
```
Kaskyi, line 27: тут

Kaskyi (approved):

packages/administration/src/admin.service.ts
Sumragen: backward compatibility

PR #1740, opened 2023-07-18
Refactor/package components fix glob path

Kaskyi:


packages/cli/src/doc-command.ts
```diff
@@ -53,7 +53,7 @@ export class DocCommand implements Command {
 
             // Get all examples inside the module
             const componentsPathMask = path.join(path.dirname(modulePath), '/**/*.component.*');
-            const componentsPath: string[] = glob.sync(componentsPathMask, { sync: true });
+            const componentsPath: string[] = glob.sync(componentsPathMask, { sync: true }).map(filePath => filePath.replace(/\\/g, '/'));
```
zverbeta, line 56: it looks like what needs to be done through a function https://nodejs.org/dist/latest-v18.x/docs/api/url.html

PR #1759, opened 2023-07-24
fix(xm-navbar-dashboard-edit-widget): open close state fixed

LevBD:
Fixed a defect when the modal for editing the dashboard constantly opened, as if not to click on the edit button again.
 How it works now:

 -when we press the button edit dashboard, the state of the open modal window is saved in the session store.

- when we press the button to close the modal for editing, this value is removed from the session store.

### DIFF
--- a/packages/administration/dashboards-config/dashboard-editor.service.ts
+++ b/packages/administration/dashboards-config/dashboard-editor.service.ts
@@ -4,12 +4,12 @@ import { SidebarRightService } from '@xm-ngx/components/sidebar-right';
 import { Dashboard, DashboardWidget } from '@xm-ngx/core/dashboard';
 import { IId } from '@xm-ngx/interfaces';
 import { XmToasterService } from '@xm-ngx/toaster';
-import { Subject } from 'rxjs';
+import { Observable, Subject } from 'rxjs';
 
 @Injectable()
 export class DashboardEditorService {
 
-    public isEdit: Subject<boolean> = new Subject<boolean>();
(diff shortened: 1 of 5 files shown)


packages/administration/dashboards-config/dashboard-editor.service.ts
```diff
@@ -4,10 +4,12 @@ import { SidebarRightService } from '@xm-ngx/components/sidebar-right';
 import { Dashboard, DashboardWidget } from '@xm-ngx/core/dashboard';
 import { IId } from '@xm-ngx/interfaces';
 import { XmToasterService } from '@xm-ngx/toaster';
+import { Subject } from 'rxjs';
 
 @Injectable()
 export class DashboardEditorService {
 
+    public isEdit: Subject<boolean> = new Subject<boolean>();
```
Sumragen, line 12: it's unsafe to make subject public. Let's make it private and create public method that returns observable instead of subject

packages/administration/dashboards-config/dashboard-editor.service.ts
```diff
@@ -17,6 +19,11 @@ export class DashboardEditorService {
 
     public close(): void {
         this.layoutService.close();
+        sessionStorage.removeItem('NAVBAR_DASHBOARD_EDIT_STORAGE_KEY');
+    }
+
+    public isEditing(isEditing?: boolean): void {
```
Sumragen, line 25: isn't clear function name. Looks like a question and as a consumer I will expect that method will return boolean.
According to method body I would probably recommend name `changeEditState`

PR #1770, opened 2023-07-27
feat: use JavascriptCode for condition in dashboard.config.hidden

Olensavchenko:


grankovsky (approved):

packages/dashboard/menu/menu.component.ts
```diff
@@ -75,12 +76,16 @@ export class MenuComponent implements OnInit, OnDestroy {
     };
 
     public ngOnInit(): void {
-        const dashboards$ = this.dashboardService.dashboards$().pipe(
-            startWith([]),
-            filter((dashboards) => Boolean(dashboards)),
-            map((i) => filterByConditionDashboards(i, this.contextService)),
-            map((i) => _.filter(i, (j) => (!j.config?.menu?.section || j.config.menu.section === 'xm-menu'))),
-            map(dashboards => buildMenuTree(dashboards)),
+        const dashboards$ = this.userService.user$().pipe(
```
denisogolubev, line 79: This piece of code should be definitely in separated method:
```ts
this.userService.user$().pipe(
            switchMap((user) => {
                return this.dashboardService.dashboards$().pipe(
                    startWith([]),
                    filter((dashboards) => Boolean(dashboards)),
                    map((i) => filterByConditionDashboards(i, this.contextService)),
                    map((i) => _.filter(i, (j) => (!j.config?.menu?.section || j.config.menu.section === 'xm-menu'))),
                    map((dashboards) => buildMenuTree(dashboards, ConditionDirective.checkCondition, {user: user})),
                );
            }),
        );
```
It's not a good practice to expand code in the Angular lifecycle hooks. 
I see that there are a lot of other code, but you, at least, can move this piece to the separated method.
So in the onInit will be one line of the code:
```ts
const dashboards$ = this.getDashboards(); // or any name you chose
```

And maybe you can play with combineLatest for this piece of code.

Olensavchenko: make separate method in the new commit

denisogolubev (approved):

PR #1774, opened 2023-08-03
chore(edit-widget): add edit widget to registry

zverbeta:
Add edit widget to registry
Text control remove required by default

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "xm-webapp",
-  "version": "5.1.7",
+  "version": "5.1.8",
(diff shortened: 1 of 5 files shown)


grankovsky (approved):

denisogolubev (approved):

oscaryald (approved):

Sumragen (changes requested):
We could not change default behaviour of xm-text-control validation, especially `required` param. Some usages are count on that default behaviour. We need to be sure that everything will work fine after those changes.

PR #1777, opened 2023-08-05
Controllers injector

Sumragen:
The following pull request provides an ability to dynamically inject services like it works with components.
In order to add dynamic services all we need is to pass `@Input` parameter into dynamic presentation directive with name `controllers` which will inject service instance into the Angular injector.
Controllers config has own structure:
```
{
   key: string, // required field in order to associate Class instance in the injector dependency tree
   selector: string, // uniq identification key that you defined in registry file, associated with service Class
   config: XmConfig, // literally any object
}
```

How to inject (TBD):
We could inject controller into component using `XmDynamicInjectionTokenStoreService` with following syntax
```
private dynamicInjectionTokenStore = inject(XmDynamicInjectionTokenStoreService);
public editStateService = inject(this.dynamicInjectionTokenStore.resolve('edit-state'));
```

`XmDynamicInjectionTokenStoreService` provides `InjectionToken` by `string` value. This token will be used as the key for controller instance.

### DIFF
--- a/packages/dynamic/presentation/xm-dynamic-presentation-base.directive.ts
+++ b/packages/dynamic/presentation/xm-dynamic-presentation-base.directive.ts
@@ -8,16 +8,14 @@ import {
     SimpleChanges,
     ViewContainerRef,
 } from '@angular/core';
-import { XmDynamicServiceFactory, XmDynamicWithSelector } from '@xm-ngx/dynamic';
-import {
-    XmDynamicInjectionTokenStoreService
-} from '@xm-ngx/dynamic/src/services/xm-dynamic-injection-token-store.service';
 import { XmConfig } from '@xm-ngx/interfaces';
 import { setComponentInput } from '../operators/set-component-input';
-import { XmDynamicSelector, XmDynamicWithConfig, XmLayoutNode } from '../src/interfaces';
+import { XmDynamicServiceFactory } from '../services/xm-dynamic-service-factory.service';
+import { XmDynamicSelector, XmDynamicWithConfig, XmDynamicWithSelector, XmLayoutNode } from '../src/interfaces';
 import { XmDynamicConstructor } from '../src/interfaces/xm-dynamic-constructor';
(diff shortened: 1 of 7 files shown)


Kaskyi (approved):

packages/dynamic/presentation/xm-dynamic-presentation-base.directive.ts
```diff
@@ -32,8 +44,14 @@ export interface XmDynamicPresentationEntryModule extends XmDynamicEntryModule<X
     entry: XmDynamicPresentationConstructor;
 }
 
+export interface XmDynamicControllerConfig extends XmDynamicWithConfig, XmDynamicWithSelector {
```
denisogolubev, line 47: Should it be in model file?

denisogolubev (approved):

LevBD (approved):

packages/controllers/features/resource-data/resource-data.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    private useCache: boolean = false;
+
+    public get(): Observable<T> {
+        if (this.useCache) {
+            return this.data$.pipe(shareReplay(1));
+        }
+
+        this.useCache = true;
+        return this.resourceController.get().pipe(
+            // map(response => response.body),
```
Kaskyi, line 29: pls remove

packages/controllers/features/resource-data/resource-data.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    public update(entity: T): Observable<T> {
+        this.data$.next(entity);
+        return this.data$;
+    }
+
+    public save(): Observable<T> {
+        if (this.data$.value?.id) {
+            return this.resourceController.update(this.data$.value);
+        }
+        //     this.resourceController.create(this.data$.value)
```
Kaskyi, line 47: pls remove

PR #1778, opened 2023-08-07
MWUSSD1-277: xm-table-widget storage and pagination fix

denisogolubev:
- Add support for the several `xm-table-widgets` on the same page, dialog, etc
- Add support for the several `xm-table-widgets` filters on the same page, dialog, etc
- Fixed pagination bug when the last element in the table was sliced

### DIFF
--- a/packages/repositories/src/i-entity-collection-pageable.ts
+++ b/packages/repositories/src/i-entity-collection-pageable.ts
@@ -25,7 +25,7 @@ export interface PageableAndSortable extends Pageable, Sortable {
 
 export const PAGEABLE_AND_SORTABLE_DEFAULT: PageableAndSortable = {
     pageIndex: 0,
-    pageSize: 0,
+    pageSize: Number.MAX_SAFE_INTEGER,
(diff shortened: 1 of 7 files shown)


Kaskyi (approved):

packages/repositories/src/i-entity-collection-pageable.ts
```diff
@@ -25,7 +25,7 @@ export interface PageableAndSortable extends Pageable, Sortable {
 
 export const PAGEABLE_AND_SORTABLE_DEFAULT: PageableAndSortable = {
     pageIndex: 0,
-    pageSize: 0,
+    pageSize: Number.MAX_SAFE_INTEGER,
```
Sumragen, line 28: In this case we will achieve an issue when collection tries to retrieve literally everything.
image

Sumragen (changes requested):

grankovsky (approved):

Sumragen (approved):

PR #1797, opened 2023-08-22
fix(xm-table): inline filter states

zverbeta:
Users may be able to config removable and disabled field in inline filters


grankovsky (approved):

oscaryald (approved):

packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -19,7 +19,16 @@ import { FormLayoutItem } from '@xm-ngx/components/form-layout';
 import { XmInlineControlConfig, XmInlineControlDynamic, XmInlineControlDynamicView } from '@xm-ngx/components/inline-control';
 import { XmTableFilterChipsControlComponent } from './xm-table-filter-chips-control.component';
 
-const DEFAULT_CONFIG: XmTableFiltersControlRequestConfig = {
+export interface XmTableInlineFilterFormLayoutItem extends FormLayoutItem {
+    removable?: boolean;
+}
+
+export interface XmTableInlineFilterChipsConfig extends XmTableFiltersControlRequestConfig {
+    filters: XmTableInlineFilterFormLayoutItem[];
```
denisogolubev, line 27: Are these arrays the same?
```ts
filters: XmTableInlineFilterFormLayoutItem[];
chips: XmTableInlineFilterFormLayoutItem[];
```

Do we really need both of them or we can just use 1 array and make 2 UI elements?

packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -226,6 +238,7 @@ export class XmTableFilterChipsComponent {
                     config: config,
                     inlineConfig: inlineConfig,
                     value: this.value[config.name],
+                    removable: config?.removable != false,
```
denisogolubev, line 241: Is there any point in doing this check `config?.removable != false`?
As I see `removable` can be only `boolean`, so this check looks like this:
`true === true`

Or am I missing something?

packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -19,7 +19,16 @@ import { FormLayoutItem } from '@xm-ngx/components/form-layout';
 import { XmInlineControlConfig, XmInlineControlDynamic, XmInlineControlDynamicView } from '@xm-ngx/components/inline-control';
 import { XmTableFilterChipsControlComponent } from './xm-table-filter-chips-control.component';
 
-const DEFAULT_CONFIG: XmTableFiltersControlRequestConfig = {
+export interface XmTableInlineFilterFormLayoutItem extends FormLayoutItem {
```
denisogolubev, line 22: All interfaces should be in the separated `model.ts` file.

packages/components/table/components/xm-table-filter-chips.component.ts
zverbeta: This request won't contain any changes except fixes

packages/components/table/components/xm-table-filter-chips.component.ts
zverbeta: You are missing case when removable will be undefined

packages/components/table/components/xm-table-filter-chips.component.ts
zverbeta: This request won't contain any changes except fixes

Sumragen (approved):

packages/components/table/components/xm-table-filter-chips.component.ts
denisogolubev: I'm not missing this case:

https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toboolean

> 2. If argument is one of undefined, null, +0𝔽, -0𝔽, NaN, 0ℤ, or the empty String, return false.

If you want to convert it to boolean use negation operator:
```ts
!!config?.removable
```

But even this is not necessary, you can keep `config?.removable` and that will be more than enough.

packages/components/table/components/xm-table-filter-chips.component.ts
denisogolubev: JFI -> You are doing these fixes out of our agreements.

packages/components/table/components/xm-table-filter-chips.component.ts
zverbeta: > !!config?.removable

value != null or value !== false  is enough

denisogolubev: JFI -> You are doing an additional check where you don't need it. So it's over-coding.

PR #1817, opened 2023-08-30
feat(xm-table): add tableEventManagerService

oscaryald:


packages/components/table/service/xm-table-event-manager.service.ts
```diff
@@ -0,0 +1,35 @@
...
+interface EventManagerAction<T = any> {
+    name: EventManagerKey;
+    payload?: T;
+}
+
+@Injectable({
+    providedIn: 'root',
+})
+export class XmTableEventManagerService {
+    private dispatcher: BehaviorSubject<EventManagerAction> = new BehaviorSubject<EventManagerAction>({name: null});
```
Kaskyi, line 17: BehaviorSubject will cache all data from all tables because it in root
i suggest to change to Subject

oscaryald: done

packages/components/table/directives/xm-table.directive.ts
```diff
@@ -89,20 +89,29 @@ export class XmTableDirective implements OnInit, OnDestroy {
...
         );
 
-        combineLatest([
-            this.tableFilterController.change$(),
-            this.pageableAndSortable$,
-        ])
-            .pipe(takeUntilOnDestroy(this))
-            .subscribe(([filterParams, pageableAndSortable]) => {
-                const queryParams = _.merge({}, { pageableAndSortable }, { filterParams });
+        const obsList$: any = [this.tableFilterController.change$(), this.pageableAndSortable$];
```
zverbeta, line 97: Using object with keys instead of array make access easier later 

Later 

combineLatestResult.tableFiler
combineLatestResult.page
combineLatestResult....

packages/components/table/service/xm-table-event-manager.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+interface EventManagerAction<T = any> {
+    name: EventManagerKey;
+    payload?: T;
+}
+
+@Injectable({
+    providedIn: 'root',
+})
+export class XmTableEventManagerService {
```
zverbeta, line 16: It looks like a duplication of this https://github.com/xm-online/xm-webapp/blob/d47e7c5317b6f9ddfa6de76e83d349744da5aed3/src/app/shared/core/src/xm-event-manager.service.ts#L18

zverbeta (approved):

Kaskyi (approved):

PR #1821, opened 2023-08-31
feature(autocomplete-control): Interfaces refactoring to make possible extend.

denisogolubev:
- Separated nested objects to the different interfaces to make possible extension


packages/components/autocomplete-control/autocomple-control.interface.ts
```diff
@@ -51,6 +42,24 @@ export interface XmAutocompleteControlConfig {
...
+    body: XmAutocompleteControlBody;
+    headers: Record<string, string>;
+}
+
+export interface IFetchAutocompleteControlParams {
+    body: XmAutocompleteControlParams;
+    queryParams: XmAutocompleteControlParams;
+}
+
+export interface IRouteQueryParams {
```
zverbeta, line 58: It might be redundant

packages/components/autocomplete-control/autocomple-control.interface.ts
```diff
@@ -51,6 +42,24 @@ export interface XmAutocompleteControlConfig {
     required?: boolean;
 }
 
+export interface ISearchAutocompleteControlParams {
```
zverbeta, line 45: Could we use XmAutocomplete prefix instead of just I

Kaskyi (approved):

LevBD (approved):

PR #1835, opened 2023-09-18
Fix redirect on invalid token

sergeysenja1992:


packages/core/auth/src/xm-authentication.service.ts
```diff
@@ -9,6 +9,7 @@ import { AuthRefreshTokenService } from './auth-refresh-token.service';
 import { GuestTokenResponse } from './guest-token.response';
 import { AuthTokenResponse } from './auth-token.response';
 import { XmAuthenticationConfig } from './xm-authentication-config.service';
+import { Router } from "@angular/router";
```
Bozonych, line 12: Pls, change double qoute to single qoute: " => '

sergeysenja1992: @Bozonych  fixed

Kaskyi (approved):

PR #1836, opened 2023-09-18
chore(xm-table): some fixes

zverbeta:
Keep fixed filters on both view
Add padding if pagination is missing

### DIFF
--- a/packages/components/table/table-widget/xm-table-widget.component.html
+++ b/packages/components/table/table-widget/xm-table-widget.component.html
@@ -7,7 +7,10 @@
               [xmTableController]="getCollectionController()"
 >
 
-  <mat-card *ngIf="xmTable.context$ | async as context" class="no-horizontal-paddings mb-3">
+  <mat-card *ngIf="xmTable.context$ | async as context" [ngClass]="{
+    'no-horizontal-paddings mb-3': true,
+    'pb-4': xmTable.config?.pageableAndSortable?.hidePagination,
+  }">
 
     <xm-table-header [config]="config" [loading]="context.collection.loading">
       <xm-table-filter-button [config]="config"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "xm-webapp",
-  "version": "5.3.22",
+  "version": "5.3.23",
   "release": "5.0.0",
   "private": true,
   "description": "Xm-webapp",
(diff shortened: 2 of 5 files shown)


packages/components/table/table-widget/xm-table-widget.component.html
```diff
@@ -7,7 +7,10 @@
               [xmTableController]="getCollectionController()"
 >
 
-  <mat-card *ngIf="xmTable.context$ | async as context" class="no-horizontal-paddings mb-3">
+  <mat-card *ngIf="xmTable.context$ | async as context" [ngClass]="{
+    'no-horizontal-paddings mb-3': true,
+    'pb-4': xmTable.config?.pageableAndSortable?.hidePagination,
```
Kaskyi, line 12: `@Defaults` already provide a value; you don't need to check the field with a '?'.

Kaskyi (approved):

grankovsky (approved):

packages/components/table/table-widget/xm-table-widget.component.html
```diff
@@ -7,7 +7,10 @@
               [xmTableController]="getCollectionController()"
 >
 
-  <mat-card *ngIf="xmTable.context$ | async as context" class="no-horizontal-paddings mb-3">
+  <mat-card *ngIf="xmTable.context$ | async as context" [ngClass]="{
+    'no-horizontal-paddings mb-3': true,
```
denisogolubev, line 11: `'no-horizontal-paddings mb-3': true` here is no logic, so it's better to use class property to assign these classes to the `mat-card`

denisogolubev (approved):

packages/components/table/table-widget/xm-table-widget.component.html
zverbeta: Agree, but also no logic have 2 different method apply a class

denisogolubev: It's not that bad, because everything has to be responsible they are created for.
`ngClass` directive is responsible for dynamically add/remove class.
`class` attribute – to assign class on element initialization.

Also it's not a good tone of code to write
```ts
if (true) {
  el.class = `${el.class} someNewClass`
}
```

because it's the same as
```ts
el.class = `${el.class} someNewClass`
```

So why is it necessary to execute additional check if you already know that these classes should be there?😊

PR #1840, opened 2023-09-21
Make idp logout configurable

sergeysenja1992:


major13ua (approved):

packages/account/logout/logout.component.ts
```diff
@@ -61,7 +61,7 @@ export class LogoutComponent implements OnInit, OnDestroy {
                 takeUntilOnDestroy(this),
             )
             .subscribe(([logout, config]) => {
-                if (idpLogoutUri && config?.idp?.enabled && idpClient) {
+                if (idpLogoutUri && config?.idp?.enabled && config?.idp?.logoutFromIdp && idpClient) {
```
zverbeta, line 64: What is the difference between enabled and logoutFromIdp?

sergeysenja1992: enabled - it`s about full idp feature, about login buttons etc
logoutFromIdp - it`s only about logout

now idp logout from google, I want by default logout from only our system

PR #1845, opened 2023-10-02
feat(core): add possibility to show data as wrapped lines

grankovsky:
add option to display list as wrapped row

### DIFF
--- a/packages/components/table/cells/table-array/table-array.component.ts
+++ b/packages/components/table/cells/table-array/table-array.component.ts
@@ -1,22 +1,32 @@
 import { CommonModule } from '@angular/common';
 import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@angular/core';
-import {
-    XmDynamicPresentation,
-} from '@xm-ngx/dynamic';
-import { defaults, get as _get, find as _find } from 'lodash';
+import { XmDynamicPresentation, } from '@xm-ngx/dynamic';
+import { defaults, find as _find, get as _get } from 'lodash';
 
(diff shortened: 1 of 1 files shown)


zverbeta (approved):

packages/components/table/cells/table-array/table-array.component.ts
```diff
@@ -53,8 +65,10 @@ export class XmTableArrayComponent implements XmDynamicPresentation<XmTableArray
             this.config?.predicate || this.config.fieldKey
         )) {
             const found = _find(this.value, this.config.predicate);
-
             this.data = _get(found, this.config.fieldKey, null);
         }
+        if (this.config?.splitLine && this.config?.splitLine?.splitSymbol && typeof (this.data) === 'string') {
```
denisogolubev, line 70: This can be written a bit easier to read with the same checks:
```ts
const { splitSymbol } = this.config?.splitLine || {};
if (splitSymbol && typeof (this.data) === 'string') {
    this.list = this.data.split(splitSymbol);
}
```

Also this condition ` || ','` in your version is unreachable:
```ts
this.list = this.data.split(this.config?.splitLine?.splitSymbol || ',');
```
because your if condition checks and executes this logic only when `splitSymbol` is specified.

Kaskyi (approved):

packages/components/table/cells/table-array/table-array.component.ts
grankovsky: updated

denisogolubev (approved):

PR #1852, opened 2023-10-09
chore(packages): fixes

zverbeta:


Bozonych (approved):

grankovsky (approved):

packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
```diff
@@ -62,6 +69,13 @@ export class XmTableSettingStore implements OnDestroy {
         delete state[key];
         this.store.next(state);
     }
+
+    private hasColumnChanges(s1: XmTableSettingStoreStateItem, s2: XmTableSettingStoreStateItem): boolean {
+        const c1 = (s1?.columns ?? []).map(s => s.name).join(',');
+        const c2 = (s2?.columns ?? []).map(s => s.name).join(',');
```
denisogolubev, line 75: code duplication
```ts
(s2?.columns ?? []).map(s => s.name).join(',');
```

Would be nice to separate it into reusable method.

packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
```diff
@@ -62,6 +69,13 @@ export class XmTableSettingStore implements OnDestroy {
...
+
+        return c1 != c2;
```
denisogolubev, line 77: Shouldn't it be a strict check `!==`?

denisogolubev (approved):

packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
zverbeta: Why?! This is not appropriate here

PR #1859, opened 2023-10-11
Add redirect after login on 401 and fix account 500

sergeysenja1992:


andrewkikot (approved):

packages/core/user/src/principal.service.ts
```diff
@@ -128,6 +129,12 @@ export class Principal implements OnDestroy, OnInitialize {
                 return;
             }
 
+            if (this.xmAuthenticationService.isSureGuestSession()) {
```
zverbeta, line 132: Identity service using this https://github.com/xm-online/xm-webapp/blob/167fe26c9dbda16b00df6eeff1351fa431cc3639/packages/core/permission/src/user-route-access-service.ts#L48

You should move this to UserRouteAccessService

packages/core/auth/src/xm-authentication.service.ts
```diff
@@ -45,6 +49,7 @@ export class XmAuthenticationService {
             tap((res) => this.updateTokens(res)),
             catchError(() => {
                 this.sessionService.clear();
+                this.xmAuthTargetUrlService.storeCurrentUrl();
```
zverbeta, line 52: You can create a custom interceptor or reuse the ErrorHandler, and store the current URL if the request fails.

sergeysenja1992: this is logout on 401 logic, and this line part of this login

packages/core/user/src/principal.service.ts
sergeysenja1992: if I follow you suggestion, i will copy paste this code in several place and break encapsulation

this actual not only for router, this is actual for any places where present identity() call

packages/core/auth/src/xm-authentication.service.ts
zverbeta: Moving this redirect logic (save urls) to the interceptor will allow us to expand it and remove the binding to the authorization logic.

packages/core/auth/src/xm-authentication.service.ts
```diff
@@ -55,6 +60,28 @@ export class XmAuthenticationService {
         return { Authorization: `Bearer ${token}` };
     }
 
+    // return true when it's guest, and return false when not sure
+    public isSureGuestSession(): boolean {
```
Kaskyi, line 64: `this.sessionService.isActive()` might be alternative

Kaskyi (commented):
pls update

packages/core/user/src/principal.service.ts
sergeysenja1992: in this case Principal will be depends on UserRouteAccessService

packages/shared/src/shared.module.ts
```diff
@@ -116,6 +117,7 @@ export class XmSharedModule {
                 RegisterService,
                 AccountService,
                 StateStorageService,
+                XmAuthTargetUrlService,
```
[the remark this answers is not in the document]
sergeysenja1992: XmAuthenticationService or XmAuthTargetUrlService?

packages/core/auth/src/xm-authentication.service.ts
sergeysenja1992: catchError contains logic that move to login page on 401 when updateTokens failed, before we move to login page we need to save current url

we don't need to save current on every http call or on every open dashboard
we don't need to `remove the binding to the authorization logic`, because saving url it`s part of this logic

packages/core/auth/src/xm-authentication.service.ts
sergeysenja1992: this.sessionService.isActive() need to be set, but I need check current token situation, and call `account` only for user session, not for client or guest

packages/shared/src/shared.module.ts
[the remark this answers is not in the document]
sergeysenja1992: fixed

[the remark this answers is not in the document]
Kaskyi: pls remove from XmSharedModule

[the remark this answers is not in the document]
Kaskyi: * XmAuthTargetUrlService

packages/core/user/src/principal.service.ts
zverbeta: Nope, UserRouteAccessService will be depend on XmAuthenticationService

packages/core/auth/src/xm-authentication.service.ts
zverbeta: It will not be added on every http call, it will be added only for errors, like here in catchError but at the global level

packages/shared/src/shared.module.ts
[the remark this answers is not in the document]
sergeysenja1992: fixed

packages/core/user/src/principal.service.ts
sergeysenja1992: this check actual for all this identity calls

image

packages/core/auth/src/xm-authentication.service.ts
sergeysenja1992: refactoring out of scope of this task

major13ua (approved):

PR #1875, opened 2023-10-26
feat(global): use npm link instead of workspaces

Kaskyi:


Kaskyi:
require test approval

PR #1882, opened 2023-10-26
refactor(example-webapp-ext): move to packages

Kaskyi:
Move example-webapp-ext to packages.
Make independent module.
Code module location is @xm-ngx/example-webapp-ext;


sergeysenja1992 (changes requested):
Why you move example of extension to packages?
Example of extensions HAVE TO leave as extension, not as packages?

sergeysenja1992:
Canceled by agreed with @Kaskyi

PR #1883, opened 2023-10-26
refactor(dashboard/home): move to packages

Kaskyi:
move home to packages


sergeysenja1992 (changes requested):
Why we need move home to packages?

sergeysenja1992 (approved):

Kaskyi:
> Why do we need to move the home to packages?

To remove all code from the /src/app folder and keep only the config file. The home serves for the login page generation. It could be replaced with custom implementation.

denisogolubev (approved):

grankovsky (approved):

PR #1906, opened 2023-11-20
fix: join value of chip to string

Olensavchenko:


LevBD (approved):

denisogolubev (approved):

Sumragen (approved):

Sumragen:
Will be better to have an ability to specify amount of array items to display. 
For example in case when we choose 10 items single chip could show us "1 item, 2 item (+8 more)"

PR #1913, opened 2023-11-29
feat: add translations and separate chips

Olensavchenko:


packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -149,6 +149,9 @@ export interface XmTableFilterInlineFilter {
         MatIconModule,
         MatMenuModule,
         XmTableFilterChipsControlComponent,
+        XmTranslationModule,
```
denisogolubev, line 152: 3 times duplicated import

packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -193,7 +196,11 @@ export class XmTableFilterChipsComponent {
 
     public remove(filter: XmTableFilterInlineFilter): void {
         const copy = cloneDeep(this.value);
-        delete copy[filter.name];
+        if(isArray(copy[filter.name])){
+            copy[filter.name] = (copy[filter.name] as any[]).filter(value => value !== filter.value)
```
denisogolubev, line 200: According to the type you specified in the interface `FiltersControlValue` here should be `Primitive[]` not `any[]`

packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -238,19 +245,34 @@ export class XmTableFilterChipsComponent {
                     view: overrideView as XmInlineControlDynamicView<unknown, unknown>,
                     edit: config as XmInlineControlDynamic<unknown>,
                 };
+                if(isArray(this.value[config.name])){
+                    return (this.value[config.name] as any[]).map((value, index) => {
```
denisogolubev, line 249: According to the type you specified in the interface FiltersControlValue here should be Primitive[] not any[]

packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -238,19 +245,34 @@ export class XmTableFilterChipsComponent {
...
+                        let title;
+                        if(config['title']) {
+                            title = config['title'][value] ? config['title'][value] : config.name;
+                        } else {
+                            title = config['title'] || config.name;
+                        }
+                        return {
```
denisogolubev, line 256: if the key and value have the same name you can write only the value, then this `return` will be simpler, but do the same thing:
```ts
return {
     config,
     inlineConfig,
     value,
     removable: !config?.removable,
     title,
     name: config.name,
 } as XmTableFilterInlineFilter;
```

packages/components/table/components/xm-table-filter-chips.component.ts
```diff
@@ -238,19 +245,34 @@ export class XmTableFilterChipsComponent {
...
+                            inlineConfig: inlineConfig,
+                            value: value,
+                            removable: config?.removable !== false,
+                            title: title,
+                            name: config.name,
+                        } as XmTableFilterInlineFilter;
+                    })
+                }
+
                 return {
```
denisogolubev, line 267: also, this second return looks the same as the previous one, with some differences.

To not duplicate the same rows you can create the variable where will be stored initial data (duplicated one)
```ts
 const tableFilterInlineFilter: XmTableFilterInlineFilter = {
     config,
     inlineConfig,
     removable: !config?.removable,
     name: config.name,
 }
// then the first return will be like this
 return {
     ...tableFilterInlineFilter,
     value,
     title
 }

// and the second return will be like this
 return {
     ...tableFilterInlineFilter,
     value: this.value[config.name],
     title: config['title'] || this.value[config.name],
 }
```

in this way less duplicated code, and you can easily make changes for both of these returns.
But, these changes are up to you. Just an alternative way to write the code.

LevBD (approved):

denisogolubev (approved):

Sumragen (approved):

PR #1928, opened 2023-12-12
fear(xm-table-filter-inline): add xm-table-filter-inline

oscaryald:
add params to config popUpFilter, showFilterChips
add xm-table-filter-inline
fix in XmTableFilterButtonDialogControlComponent setting value to form,
made inline filter by default


packages/components/table/components/xm-table-filter-inline.component.ts
```diff
@@ -0,0 +1,115 @@
...
+    ],
+})
+export class XmTableFilterInlineComponent implements OnInit, OnDestroy {
+    @Input() public config: XmTableFiltersControlRequestConfig;
+    @Input() public loading: boolean;
+
+    public value: FiltersControlValue;
+    private cacheFilters: FiltersControlValue;
+    private DELAY = 400;
+    private request$: BehaviorSubject<FiltersControlValue> = new BehaviorSubject<FiltersControlValue>(null);
```
denisogolubev, line 64: Why do we need this `BehaviorSubject` here?
I see only one usage `this.request$.next(this.value);`
https://github.com/xm-online/xm-webapp/pull/1928/files#diff-9ce6e00ebaa031d2042fd6f8e2e1a8622976f3d93ad5514f6a6e095700845122R85

Why we can't just do there `this.tableFilterController.set(this.value);` (call the `submit()` function) without any additional observers?

denisogolubev (approved):

Kaskyi (approved):

zverbeta (approved):

PR #1940, opened 2023-12-27
fix: get offset from user

Olensavchenko:


packages/translation/pipes/xm-date-time.pipe.ts
```diff
@@ -41,7 +41,7 @@ export class XmDateTimePipe implements PipeTransform {
     }
 
     private getOffset(): string {
-        return this.account?.timeZoneOffset || '';
+        return this.principal.getTimezoneOffset() || '';
```
Bozonych, line 44: Ir`s overhead. getTimezoneOffset() - now returns empty string.

Bozonych (approved):

packages/core/user/src/principal.service.ts
```diff
@@ -266,6 +266,11 @@ export class Principal implements OnDestroy, OnInitialize {
         return moment().format('Z');
     }
 
+    public getTimezoneOffset(): string {
+        // For now setting offset from browser
+        return this.userIdentity.timeZoneOffset || '';
```
Bozonych, line 271: Maybe make sense check this.userIdentity**?**.timeZoneOffset ??

packages/translation/pipes/xm-date-time.pipe.ts
Olensavchenko: overhead deleted

packages/core/user/src/principal.service.ts
Olensavchenko: Yes. It makes sense.

Kaskyi (approved):

LevBD (approved):

PR #1948, opened 2024-01-11
fix(core): update resourse data/ add possibility to update by querypa…

grankovsky:
update resourse data/ add possibility to update by queryparams/ add config to select params for update


Bozonych (approved):

packages/controllers/features/resource-data/resource-data.service.ts
```diff
@@ -54,4 +75,19 @@ export class ResourceDataService<T extends IId = any> {
         this.data$.next(cloneDeep(this.stable));
         return this.data$;
     }
+
+    private shouldFilterParams(params: any): boolean {
```
denisogolubev, line 79: any?

packages/controllers/features/resource-data/resource-data.service.ts
```diff
@@ -54,4 +75,19 @@ export class ResourceDataService<T extends IId = any> {
...
+        if (this.config?.updateEntityChangeParams) {
```
denisogolubev, line 80: 3 times access to the `this.config?.updateEntityChangeParams`, could be re-written to (use de-structurization):
```ts
const {updateEntityChangeParams} = this.config || {};
if (updateEntityChangeParams) {
    const param = params[updateEntityChangeParams[0]];
    const hasIdChanged: boolean = param !== this.previousId;
    this.previousId = param;
    return hasIdChanged;
}
return true;
```

packages/controllers/features/resource-data/resource-data.service.ts
```diff
@@ -54,4 +75,19 @@ export class ResourceDataService<T extends IId = any> {
...
+            const hasIdChanged: boolean = params[this.config?.updateEntityChangeParams[0]] !== this.previousId;
+            this.previousId = params[this.config?.updateEntityChangeParams[0]];
+            return hasIdChanged;
+        }
+        return true;
+    }
+}
+
+interface DataResourceOptions{
```
denisogolubev, line 89: Interfaces should be in the `model` files

packages/controllers/features/resource-data/resource-data.service.ts
```diff
@@ -27,15 +32,31 @@ export class ResourceDataService<T extends IId = any> {
...
+                        this.data$.next(data);
+                        this.stable = cloneDeep(data);
+                        return this.data$.pipe(shareReplay(1));
+                    }),
+                );
             }),
         );
     }
 
+    private hasRelevantParamChanges(currentParams: Params, relevantParams: string[]): boolean {
```
denisogolubev, line 50: You can use `some` iterator for this case:
```ts
relevantParams?.length && relevantParams.some((key: string) => {
    this.previousParams = {...currentParams};
    return this.shouldFilterParams(currentParams) && currentParams[key] !== this.previousParams[key];
});
```

But, you can keep it as it is. Just an alternative way to write this logic.

packages/controllers/features/resource-data/resource-data.service.ts
grankovsky: fixed

packages/controllers/features/resource-data/resource-data.service.ts
grankovsky: fixed

packages/controllers/features/resource-data/resource-data.service.ts
grankovsky: replased

packages/controllers/features/resource-data/resource-data.service.ts
grankovsky: let it as it is

denisogolubev (approved):

zverbeta (approved):

PR #1955, opened 2024-01-18
feat: move key to config, get controller by key, change model

Olensavchenko:
make ResourceArrayDataService ruesable

### DIFF
--- a/packages/controllers/features/resource-array-data/resource-array-data.service.ts
+++ b/packages/controllers/features/resource-array-data/resource-array-data.service.ts
@@ -1,15 +1,13 @@
-import { Injectable } from '@angular/core';
-import { RestRepositoryService } from '@xm-ngx/controllers/features/repository/rest-repository';
-import { injectByKey } from '@xm-ngx/dynamic';
+import { inject, Injectable, Injector, ProviderToken } from '@angular/core';
+import { XmDynamicInjectionTokenStoreService } from '@xm-ngx/dynamic';
 import { IId } from '@xm-ngx/interfaces';
 import { cloneDeep, get, isEmpty } from 'lodash';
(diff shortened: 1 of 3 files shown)


packages/controllers/features/resource-array-data/resource-array-data.service.ts
```diff
@@ -29,10 +29,10 @@ export class ResourceArrayDataService<T extends IId = any> {
         }
 
         this.useCache = true;
-        return this.resourceController.get().pipe(
+        return this.getControllerByKey<RestRepositoryService>(this.config?.dataController?.key || 'data').get().pipe(
```
denisogolubev, line 32: duplicated code, you can create get method and reuse it where you need the same logic
```ts
private get key(): string {
  return this.config?.dataController?.key || 'data';
}
```

packages/controllers/features/resource-array-data/resource-array-data.service.ts
```diff
@@ -50,7 +50,7 @@ export class ResourceArrayDataService<T extends IId = any> {
 
     public save(): Observable<T> {
         if (this.data$.value?.id) {
-            return this.resourceController.update(this.data$.value).pipe(
+            return this.getControllerByKey<RestRepositoryService>(this.config?.dataController?.key || 'data').update(this.data$.value).pipe(
```
denisogolubev, line 53: duplicated code, see the previous comment.

packages/controllers/features/resource-array-data/resource-array-data.service.ts
Olensavchenko: done

oscaryald (approved):

denisogolubev (approved):

PR #1974, opened 2024-01-29
fix(core): table query-params optimization

grankovsky:
Changed queryparams flatten object. Optimize url for page with table


Bozonych (approved):

packages/components/table/controllers/filters/xm-table-query-params-store.service.ts
```diff
@@ -25,18 +25,33 @@ export class XmTableQueryParamsStoreService {
...
-                    [this.key]: JSON.stringify(queryParams)
+                    ...flattenedParams,
                 },
-                queryParamsHandling: 'merge'
+                queryParamsHandling: 'merge',
             },
         );
+    }
 
+    private flattenJson(data: QueryParamsPageable, parentKey: string = '', result: FlattenedType = {}): FlattenedType {
```
denisogolubev, line 43: This function can't do the same result?
https://lodash.com/docs/4.17.15#flatMap

zverbeta (approved):

PR #1975, opened 2024-01-29
Fix/timeline list component

LevBD:
fix get all timeline items using loadAllTimeLineItemsV2() method

### DIFF
--- a/packages/timeline/timeline-list/timeline-list.component.ts
+++ b/packages/timeline/timeline-list/timeline-list.component.ts
@@ -81,6 +81,7 @@ export class TimelineListComponent implements OnInit, OnDestroy {
     public formFilter: any = {};
     private totalCount: number = 0;
     private page: number = 0;
+    private timeLineItems = [];
 
     constructor(private timelineService: TimelineService,
                 private i18nNamePipe: I18nNamePipe,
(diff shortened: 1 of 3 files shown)


packages/timeline/timeline-list/timeline-list.component.ts
```diff
@@ -153,15 +154,42 @@ export class TimelineListComponent implements OnInit, OnDestroy {
             });
     }
 
-    public loadAllV2(): void {
-        this.timelineService.searchV2(this.getSearchBodyV2({ size: this.totalCount, page: 0 }))
+    public loadAllTimeLineItemsV2(): void {
+        this.isMoreLoading = true;
+        this.isLoading = true;
+
+        const searchAllTimeLineItemsV2 = (size: number = 1000): Observable<any> => this.timelineService
```
denisogolubev, line 161: Is it necessary to create the function inside the other function? You can create separate function and call it here:
```ts
private searchAllTimeLineItemsV2<D>(size: number = 1000): Observable<D> {
        return this.timelineService
            .searchV2(this.getSearchBodyV2({ size: size, page: 0 }))
            .pipe(
                map((resp: HttpResponse<any>) => {
                    this.totalCount -= 1000;
                    this.nextV2 = (this.totalCount - resp.body.length);

                    return resp;
                }),
                switchMap((resp) => {
                    this.timeLineItems.push(...resp.body);

                    if (this.totalCount > 1000) {
                        return this.searchAllTimeLineItemsV2(1000);
                    }

                    if (this.totalCount > 0 && this.totalCount <= 1000) {
                        return this.searchAllTimeLineItemsV2(this.totalCount);
                    }

                    return of(this.timeLineItems);
                }),
            );
    }
```

Bozonych (approved):

packages/timeline/timeline-list/timeline-list.component.ts
LevBD: Done

Olensavchenko (approved):

denisogolubev (approved):

PR #1979, opened 2024-02-02
fix(dashboard-widget-style): add dashboard widgets style

zverbeta:


Kaskyi (approved):

packages/dashboard/src/components/dashboard/dashboard.component.html
```diff
@@ -1,64 +1,58 @@
...
+      }">
     </ng-template>
-  </ng-container>
+  }
 </ng-template>
 
-<ng-container *ngIf="childrenDashboards?.length > 0 then childDashboards else emptyData"></ng-container>
-
-<ng-template #childDashboards>
-  <div fxLayout="row wrap" fxLayoutGap="16px grid">
```
sergeysenja1992, line 42: pls keep dashboard style
so many widgets rely on these styles

packages/dashboard/src/components/dashboard/dashboard.component.html
```diff
@@ -1,64 +1,58 @@
...
-    <ng-container *ngFor="let childDashboard of childrenDashboards">
-      <div *permitted="childDashboard.permission || 'DASHBOARD.GET_LIST'"
-           fxFlex="33%" fxFlex.sm="50%" fxFlex.xs="100%">
+@if (childrenDashboards?.length > 0) {
+  <div class="row g-2">
+    @for (childDashboard of childrenDashboards; track index; let index = $index) {
+      <div class="col-6 col-md-4 col-xl-2" *permitted="childDashboard?.config?.permission || 'DASHBOARD.GET_LIST'">
```
sergeysenja1992, line 43: was `childDashboard.permission`, now `childDashboard?.config?.permission` ? O_o

Sumragen (approved):

packages/dashboard/src/components/dashboard/dashboard.component.html
zverbeta: Interesting, because our interfaces say that permission should be inside confighttps://github.com/xm-online/xm-webapp/blob/main/packages/core/dashboard/models/dashboard.model.ts#L7

packages/dashboard/src/components/dashboard/dashboard.component.html
zverbeta: FlexLayout is deprecated https://github.com/angular/flex-layout so i replace it bootstrap classes, that produce same result

packages/dashboard/src/components/dashboard/dashboard.component.html
zverbeta: @sergeysenja1992 Do you have dashboard where permission property is set on the dashboard level?

sergeysenja1992: I don't have access to the entire project of all clients.(

sergeysenja1992 (approved):

PR #1980, opened 2024-02-05
fix(xm-date-range): fix date locale on change acc settings

oscaryald:
fix date locale on change account settings

### DIFF
--- a/packages/components/date-range/xm-date-range.component.ts
+++ b/packages/components/date-range/xm-date-range.component.ts
@@ -1,15 +1,13 @@
-import { DatePipe } from '@angular/common';
 import {
     ChangeDetectionStrategy,
-    Component, inject,
+    Component,
     Input,
     ViewEncapsulation,
-    OnInit
 } from '@angular/core';
 import { XmDynamicPresentation, } from '@xm-ngx/dynamic';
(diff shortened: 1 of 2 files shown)


packages/components/date-range/xm-date-range.component.ts
```diff
@@ -22,12 +30,25 @@ export const XM_DATE_RANGE_CONFIG_DEFAULT: XmDateRangeConfig = {
...
+    private xmUserService: XmUserService = inject(XmUserService);
+
+    public ngOnInit(): void {
+        this.xmUserService.user$()
+            .pipe(
+                take(1)
+            )
+            .subscribe((user) => {
+                this.locale = user.langKey;
+            });
```
Sumragen, line 52: Looks like something similar to xmPrincipal.getLang(), isn't it?

oscaryald: changed it, thanks

sergeysenja1992 (approved):

Bozonych (approved):

zverbeta (changes requested):
I see it as a separate pipe

zverbeta (approved):

PR #1987, opened 2024-02-12
Add user context

sergeysenja1992:


YRedko (approved):

andrewkikot (approved):

packages/core/user/src/principal.service.ts
```diff
@@ -28,13 +29,20 @@ export class Principal implements OnDestroy, OnInitialize {
...
                 private userService: XmUserService,
+                private contextService: ContextService,
     ) {
     }
 
     public init(): void {
         this.sessionService.isActive().pipe(
             filter(i => i === false),
         ).subscribe(() => this.logout());
+        this.authenticationState.pipe(
```
zverbeta, line 40: It's odd, because we subscribe to the auth state and then set it as is once, bypassing the pubsub approach

zverbeta (approved):

packages/core/user/src/principal.service.ts
sergeysenja1992: What you propose?

PR #1989, opened 2024-02-12
Fix empty query params error

Sumragen:
When some of `keysToCheck` has not defined we will face with the problem of `reading *** of undefined`. Looks like if we ignore them, everything will work fine.

### DIFF
--- a/packages/components/table/controllers/filters/xm-table-query-params-store.service.ts
+++ b/packages/components/table/controllers/filters/xm-table-query-params-store.service.ts
@@ -89,7 +89,7 @@ export class XmTableQueryParamsStoreService {
         const keysToCheck: string[] = ['pageIndex', 'pageSize', 'sortBy', 'sortOrder'];
         const changedValues: PageableAndSortable = {};
         keysToCheck.forEach((key: string) => {
-            if (configParams[key].toString() !== pageableAndSortable[key].toString()) {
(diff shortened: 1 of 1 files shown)


sergeysenja1992 (approved):

packages/components/table/controllers/filters/xm-table-query-params-store.service.ts
```diff
@@ -89,7 +89,7 @@ export class XmTableQueryParamsStoreService {
         const keysToCheck: string[] = ['pageIndex', 'pageSize', 'sortBy', 'sortOrder'];
         const changedValues: PageableAndSortable = {};
         keysToCheck.forEach((key: string) => {
-            if (configParams[key].toString() !== pageableAndSortable[key].toString()) {
+            if (configParams[key] && pageableAndSortable[key] && configParams[key].toString() !== pageableAndSortable[key].toString()) {
```
denisogolubev, line 92: Could be simplified 
```ts
if (configParams[key]?.toString() !== pageableAndSortable[key]?.toString())
```

Sumragen: By some point of view, you are right, but not completely. The reason why is that `configParams[key]` and `pageableAndSortable[key]` could have different values - `undefined` or `null`.
image

But thanks to you, I see another weakness of that solution - `configParams[key]` or `pageableAndSortable[key]` could have value like `0`, `false`, etc. Due of that it's better to check whenever that value is present.

PR #1992, opened 2024-02-13
fix: set pageIndex equal 0 after clicking sort or filter

Olensavchenko:


LevBD (approved):

packages/components/table/directives/xm-table.directive.ts
```diff
@@ -120,6 +121,10 @@ export class XmTableDirective implements OnInit, OnDestroy {
             .subscribe((obsObj) => {
                 const filterParams = obsObj.tableFilter;
                 const pageableAndSortable = obsObj.pageableAndSortable;
```
denisogolubev, line 123: This piece of code can be separated into different method to not expand the subscription's logic.
```ts
// inside subscription
const pageableAndSortable = this.mapPagebleAndSortable(filterParams, obsObj.pageableAndSortable);

// separate method for mapping
private mapPagebleAndSortable(filterParams, pageableAndSortable) {
  if (!isEqual(filterParams, this.filters)){
      set(pageableAndSortable, 'pageIndex', 0);
   }
  return pageableAndSortable;
}
```
In this way, you group specific parts of your code into methods, and in subscription, we can read only the sequence of execution but not logic.

denisogolubev (approved):

Bozonych (approved):

PR #1995, opened 2024-02-14
feat(inject-key-type): inject key or type

zverbeta:


Kaskyi (approved):

sergeysenja1992 (approved):

Sumragen (changes requested):
there is no need to mix "native" angular API with our specific. Otherwise this method is no longer need to have mention "byKey" in name.

PR #2000, opened 2024-02-22
Add resource url interpolation for the table component

denisogolubev:
- add possibility to config interpolation params and get them from url query params


packages/components/table/repositories/xm-entity-repository.service.ts
```diff
@@ -64,7 +77,13 @@ export class XmEntityRepository<T extends XmEntity>
...
 
     protected override resourceUrl(): string {
-        return this.config.resourceUrl;
+        let url: string = this.config.resourceUrl;
+        if (this.config.resourceUrlInterpolation) {
+            const interpolation: Record<string, string> =
+                xmFormatJs(this.config.resourceUrlInterpolation, this.activatedRoute?.snapshot?.queryParams || {});
+            url = interpolate(url, interpolation);
+        }
+        return url || this.config.resourceUrl;
```
zverbeta, line 86: Condition above works, any reason for this?

oscaryald (approved):

packages/components/table/repositories/xm-entity-repository.service.ts
denisogolubev: removed, no reason, tnx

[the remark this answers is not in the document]
denisogolubev: Yeah, I've simplified it.
In this way, we even don't need to specify the params we want to interpolate. Tnx)

zverbeta (approved):

PR #2022, opened 2024-03-07
feat(xm-table): add controllers, getControllerByKey, elastic-request-…

oscaryald:
add controllers, getControllerByKey into service, elastic-request-builder add createDefaultFiltersToRequest

### DIFF
--- a/packages/dynamic/src/services/xm-dynamic-injection-token-store.service.ts
+++ b/packages/dynamic/src/services/xm-dynamic-injection-token-store.service.ts
@@ -1,26 +1,15 @@
-import { inject, Injectable, InjectionToken, Injector, ProviderToken } from '@angular/core';
+import { Injectable, InjectionToken } from '@angular/core';
 import { XmDynamicService } from '../../services/xm-dynamic-service-factory.service';
 
 
 @Injectable()
 export class XmDynamicInjectionTokenStoreService {
-
     private mapper: Record<string, InjectionToken<unknown>> = {};
-
-    private injector = inject(Injector);
 
     public resolve<T extends XmDynamicService>(key: string): InjectionToken<T> {
         if (!this.mapper[key]) {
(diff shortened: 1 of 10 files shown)


packages/dynamic/src/services/xm-dynamic-injection-token-store.service.ts
```diff
@@ -7,11 +7,20 @@ export class XmDynamicInjectionTokenStoreService {
...
         return this.mapper[key];
     }
 
+    public getControllerByKey(key: string): any {
+        if(!key) {
+            return;
+        }
+        const providerToken: ProviderToken<any> = this.resolve(key);
+        return this.injector.get(providerToken, undefined,{optional: true});
+    }
```
Sumragen, line 25: DynamicInjectionTokenStoreService responsible only for injection tokens.
It's possibly the better way to have a separate class that will extend components and services.
something like:
```
export class DynamicInstance {

    protected injector = inject(Injector);
    private injectionTokenService = inject(XmDynamicInjectionTokenStoreService);

    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore
    protected getControllerByKey<T = any>(key: string, options?: InjectOptions): T {
        const providerToken: ProviderToken<any> = this.injectionTokenService.resolve(key);
        return this.injector.get(providerToken, undefined, options);
    }
}

```

Sumragen (changes requested):

packages/dynamic/src/services/xm-dynamic-injection-token-store.service.ts
oscaryald: created separate XmDynamicInstanceService

Sumragen (approved):

Bozonych (approved):

denisogolubev (approved):

Olensavchenko (approved):

PR #2035, opened 2024-03-15
add navigation class to html, XM1-35201

mak7iliano:


major13ua (approved):

src/app/layouts/navigation-class.directive.ts
```diff
@@ -0,0 +1,32 @@
...
+        this.className = val;
+    }
+
+    constructor(
+        protected router: Router,
+    ) {
+        this.router.events.subscribe(val => {
+            if (val instanceof NavigationEnd) {
+                this.className = val.url.slice(1)
+                    .replace(/[/]/g, '-')
```
sergeysenja1992, line 26: how about special chars, better to use _.words(slug).join('-')

src/app/layouts/navigation-class.directive.ts
```diff
@@ -0,0 +1,32 @@
...
+                    .replace(/\./g, '')
+                    .toLowerCase() || 'root-page';
```
sergeysenja1992, line 28: maybe better to add prefix like slug- to avoid class collision 
slug-dashboard-order
and root page can be slug-root

sergeysenja1992 (approved):

andrewkikot (approved):

PR #2039, opened 2024-03-20
fix(xm-table-filter-chips): displaying chips items

LevBD:
fix displaying chips items
Знімок екрана 2024-03-20 о 15 46 54

now the width of the container with chips is calculated taking margins

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "xm-webapp",
-  "version": "6.0.61",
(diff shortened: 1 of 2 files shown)


Kaskyi (approved):

zverbeta (changes requested):
Magic number

zverbeta (approved):

denisogolubev (approved):

PR #2041, opened 2024-03-21
Fix maintenance and app init

sergeysenja1992:


major13ua (approved):

packages/components/maintenance/maintenance.service.ts
```diff
@@ -17,13 +17,10 @@ export class MaintenanceService {
...
     public init(): void {
-        this.themeLoader.loaded$.pipe(
-            tap(() => this.applicationConfigService.setResolved(true)),
-            catchError((err) => {
-                this.setMaintenanceProgress(true);
-                return throwError(err);
-            }),
-        ).subscribe();
+        this.themeLoader.loaded$.subscribe({
+            next: (it) => this.applicationConfigService.setResolved(!!it),
```
Kaskyi, line 21: why do we need '!!'?  - it's a <boolean> already

packages/components/maintenance/maintenance.service.ts
```diff
@@ -17,13 +17,10 @@ export class MaintenanceService {
...
+            error: () => this.setMaintenanceProgress(true)
+        });
```
Kaskyi, line 23: `return throwError(err);` - it will show the error in the console,
 is it print error in the console ?

Kaskyi (approved):

andrewkikot (approved):

PR #2044, opened 2024-03-22
Visual constructor

grankovsky:
add test resize widget, and demo replace

### DIFF
--- a/packages/dashboard/src/components/dashboard/dashboard.component.html
+++ b/packages/dashboard/src/components/dashboard/dashboard.component.html
@@ -1,6 +1,6 @@
 <xm-loader [showLoader]="showLoader"></xm-loader>
 
-<div [class]="dashboard?.layout?.class" [style]="dashboard?.layout?.style">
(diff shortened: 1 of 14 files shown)


packages/dashboard/src/components/dashboard/dashboard.component.html
```diff
@@ -56,3 +59,14 @@
 } @else {
   <no-data [show]="!dashboard?.layout?.grid?.length"></no-data>
 }
+
+
+<div *ngIf="isEdit &&visualdashboard &&injector&&spec">
+  <div class="editor">
+  <xm-visual-constructor [parentItem]="{content:visualdashboard}"
```
sergeysenja1992, line 66: missed component in commit?

packages/dashboard/src/components/dashboard/dashboard.component.html
```diff
@@ -56,3 +59,14 @@
...
+                         [dashboard]="dashboard"
+                         [injector]="injector"
+                         [spec]="spec" >
+  </xm-visual-constructor>
```
sergeysenja1992, line 70: how about constructor as dynamic component?
Ho give ability to use different constructors?

PR #2048, opened 2024-03-27
Fix: Sidebar close after logout

denisogolubev:
- hide sidebar after user logout to hide unnecessary margin space reserving

### DIFF
--- a/packages/components/login/login.service.ts
+++ b/packages/components/login/login.service.ts
@@ -1,18 +1,17 @@
 import { Inject, Injectable } from '@angular/core';
 import { Params, Router } from '@angular/router';
 import { Observable, of } from 'rxjs';
-import { AuthServerProvider } from '@xm-ngx/core/user';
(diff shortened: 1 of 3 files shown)


sergeysenja1992 (approved):

packages/components/login/login.service.ts
```diff
@@ -133,12 +135,14 @@ export class LoginService {
     public logout(): void {
         this.principal.logout();
         this.authRefreshTokenService.clear();
+        this.menuService.sidenav.close();
```
zverbeta, line 138: LoginService not suppose to depend on MenuService

zverbeta (changes requested):

packages/components/login/login.service.ts
denisogolubev: Fixed. Thanks)

zverbeta (approved):

PR #2050, opened 2024-03-28
fix(redirect-url): navigate once

Sumragen:
Redirect user to previous url once logged in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "xm-webapp",
-  "version": "6.0.73",
+  "version": "6.0.77",
   "release": "6.0.0",
   "private": true,
   "description": "Xm-webapp",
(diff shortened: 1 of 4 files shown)


denisogolubev:
The reason why it do it twice
https://github.com/xm-online/xm-webapp/blob/20c504f8e1772c156f452f1ac0a2d51dcbac4832/packages/core/permission/src/user-route-access-service.ts#L63

if we remove this redirect, the bug when user after auth routes to the blank page will be fixed.
Then the setTimeout will be unnecessary.

But I didn't have enough time to retest all case and check if it impacts to the other functionality.
Maybe this info is useful for you.

Bozonych (approved):

Sumragen:
@denisogolubev For some reason I could agree with you, but just in context of "redundant redirect". `setTimeout` resolves another issue, especially dashboard one.
In case when you remove line that you mention above user will face with incorrect system behavior after redirecting to some specific page while session is not present, because system will stuck on empty page.
And one more time, as it was said in comment before, `setTimeout` resolves case when user tries to login immediately after logout, because due to lack of time for dashboard subscription we cannot retrieve fresh data according to user session.

denisogolubev (approved):

PR #2062, opened 2024-04-03
Parse query params from redirect url

Sumragen:
Previously we forgot to separate query params from redirect url


Olensavchenko (approved):

zverbeta (changes requested):
Better to use router.parseUrl

Sumragen:
@zverbeta Thanks! That approach is much more better

zverbeta (approved):

denisogolubev (approved):

PR #2086, opened 2024-04-23
fix: Make M3 menu category mapping more sustainable

denisogolubev:


grankovsky (approved):

oscaryald (approved):

packages/components/menu/menu.service.ts
```diff
@@ -159,6 +158,13 @@ export class MenuService {
         });
     }
 
+    private checkCategoryExistence(categoryKey: string): MenuCategory {
+        if (!categoryKey || !this.categories.hasOwnProperty(categoryKey)) {
+            return this.otherCategory;
+        }
+        return JSON.parse(JSON.stringify(this.categories[categoryKey])) as MenuCategory;
```
Bozonych, line 165: make sense use lodash cloneDeep 
https://www.measurethat.net/Benchmarks/Show/5306/0/lodash-clonedeep-vs-jsonparsestringify-vs-recursivecopy
because cloneDeep is fastest

denisogolubev: Thanks, that is really worth it)

PR #2091, opened 2024-04-25
fix(core): fix revert broken changes view state on create

grankovsky:
fix revert broken changes view state on create

### DIFF
--- a/packages/controllers/features/edit-state-store/edit-state-store.service.ts
+++ b/packages/controllers/features/edit-state-store/edit-state-store.service.ts
@@ -7,7 +7,7 @@ import { PageChangesStore, PageChangesStoreType } from '@xm-ngx/core/dashboard';
 @Injectable()
 export class EditStateStoreService {
     private pageChangesStore= inject(PageChangesStore);
-    private state: BehaviorSubject<EDIT_STATE> = new BehaviorSubject<EDIT_STATE>(EDIT_STATE.VIEW);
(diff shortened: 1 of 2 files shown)


packages/controllers/features/edit-state-store/edit-state-store.service.ts
```diff
@@ -7,7 +7,7 @@ import { PageChangesStore, PageChangesStoreType } from '@xm-ngx/core/dashboard';
 @Injectable()
 export class EditStateStoreService {
     private pageChangesStore= inject(PageChangesStore);
-    private state: BehaviorSubject<EDIT_STATE> = new BehaviorSubject<EDIT_STATE>(EDIT_STATE.VIEW);
+    private state: BehaviorSubject<EDIT_STATE> = new BehaviorSubject<EDIT_STATE>(EDIT_STATE.EDIT);
```
denisogolubev, line 10: It means that in every place where this state used it will be in `edit` mode initially. Is this ok for all cases?

Olensavchenko (approved):

packages/controllers/features/edit-state-store/edit-state-store.service.ts
grankovsky: yes for now pages without entity open for view/ not for edit/create

denisogolubev (approved):

LevBD (approved):

PR #2120, opened 2024-06-14
feat: Add xm-html-value component and update existing html components

denisogolubev:
- make. all component standalone
- update storybook
- add new component wrapper that can pass the value from API response `@Input() value;`
- changeDetection.OnPush for all HTML components

image
image


zverbeta (approved):

oscaryald (approved):

Bozonych (approved):

packages/components/html/html.model.ts
```diff
@@ -0,0 +1,7 @@
+import { Translate } from '@xm-ngx/translation';
+
+export interface XmHtml {
+    class?: string;
+    style?: string;
+    html: Translate;
```
Bozonych, line 6: The title is a bit confusing.

PR #2124, opened 2024-06-19
fix: add custom date adapter

Olensavchenko:


Bozonych (approved):

packages/components/date/shared/custom-date-adapter.ts
```diff
@@ -0,0 +1,22 @@
+import { NativeDateAdapter } from '@angular/material/core';
+
+export class CustomDateAdapter extends NativeDateAdapter {
+    public parse(value: any): Date | null {
+        if (typeof value === 'string' && value.indexOf('.') > -1) {
```
zverbeta, line 5: Should we use dayjs library for stuff like this

zverbeta (changes requested):

LevBD (approved):

packages/components/date/shared/custom-date-adapter.ts
Olensavchenko: will be changed in next pr

zverbeta (commented):
We will fix it in next pr

zverbeta (approved):

PR #2143, opened 2024-07-09
feat: set monday as first day in custom-date-adapter.ts

Olensavchenko:


packages/components/date/shared/custom-date-adapter.ts
```diff
@@ -17,4 +17,8 @@ export class CustomDateAdapter extends NativeDateAdapter {
         const year = date.getFullYear();
         return `${day}.${month}.${year}`;
     }
+
+    public getFirstDayOfWeek(): number {
```
denisogolubev, line 21: it looks like a getter or variable that should be 
```ts
// variable
public mondayIndex: number = 1;

// or
public get mondayIndex(): number {
  return 1;
}
```

I think such naming is a bit better than keep comments in code.

denisogolubev (approved):

packages/components/date/shared/custom-date-adapter.ts
Olensavchenko: used variable

Bozonych (approved):

x3mlestat (approved):

PR #2153, opened 2024-07-24
feat(resource-data): skip cache

zverbeta:


packages/controllers/features/resource-data/resource-data.service.ts
```diff
@@ -30,7 +30,8 @@ export class ResourceDataService<T extends IId = any> {
             return this.data$.pipe(shareReplay(1));
         }
 
-        this.useCache = true;
+        this.useCache = this.config?.skipCache ? false : true;
```
denisogolubev, line 33: You can make this code simpler. Your IDE should underline this code as a code that should be simplified.

zverbeta: ?! This line already simple.

Bozonych (approved):

packages/controllers/features/resource-data/resource-data.service.ts
denisogolubev: And if remember correctly, when you have several resource-data elements that should take info from 1 API that code can trigger the same API call several times.

zverbeta: Correct. Because of that you can disable cache through config prop, default behavior doesn't change

denisogolubev: It's not obvious behavior, so I propose to add comment in code that it could trigger such behavior to make it easier for understand for those who will use this option.

denisogolubev (approved):

oscaryald (approved):

LevBD (approved):

PR #2160, opened 2024-07-29
fix(multiple-enum): fix clear value

Olensavchenko:


grankovsky (approved):

packages/components/enum/multiple-control/xm-multiple-enum-control.component.ts
```diff
@@ -153,13 +156,12 @@ export class XmMultipleEnumControl
     }
 
     private setInitValue(): void {
```
denisogolubev, line 158: You can simplify code via removing one if statement:
```ts
const {initValue} = this.config;

this.value = initValue?.length ? initValue : null;
this.control.patchValue(this.value);
```

denisogolubev (approved):

x3mlestat (approved):

Bozonych (approved):

PR #2230, opened 2024-10-30
fix(core): clear local storage table on logout

grankovsky:


packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
```diff
@@ -30,12 +31,19 @@ export class XmTableSettingStore implements OnDestroy {
...
         const state: XmTableSettingStoreState = this.localStorage.retrieve(COLUMNS_SETTING_STORE_NAME) || {};
         this.store = new BehaviorSubject<XmTableSettingStoreState>(state);
         this.store.subscribe(value => {
             this.localStorage.store(COLUMNS_SETTING_STORE_NAME, value);
         });
+
+        this.eventManager.listenTo('USER-LOGOUT').subscribe((): void => {
+            this.localStorage.clear(COLUMNS_SETTING_STORE_NAME);
+            this.store.next({});
+            this.store.value;
```
denisogolubev, line 45: remove this line

packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
```diff
@@ -30,12 +31,19 @@ export class XmTableSettingStore implements OnDestroy {
 
     constructor(
         private localStorage: LocalStorageService,
+        protected eventManager: XmEventManager,
```
denisogolubev, line 34: I think it's not necessary to use `protected` here, use `private` instead

denisogolubev (approved):

LevBD (approved):

PR #2231, opened 2024-10-30
fix(core): fix find permissions

grankovsky:
check permissions hasPrivilegesBy  change to array


packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
```diff
@@ -144,6 +141,16 @@ export class XmTableColumnsSettingStorageService {
         );
     }
 
+    private checkPermissions(permission: string | string[], permissionStrategy: string): Observable<boolean> {
+        const permissions: string[] = Array.isArray(permission) ? permission : [permission];
+        return this.permissionService.hasPrivilegesBy(permissions, permissionStrategy as PermissionCheckStrategy)
+            .pipe(
+                tap((res)=>{
```
denisogolubev, line 148: What the purpose of this `tap`?

LevBD (approved):

zverbeta (changes requested):
What is the purpose of the Mr? I don't see any differences, could you explain more?

zverbeta (approved):

PR #2238, opened 2024-11-12
fix/nubmer-control: fix number control value

zverbeta:


packages/components/number-control/xm-number.directive.ts
```diff
@@ -59,6 +59,7 @@ export class XmNumberDirective {
         this.elementRef.nativeElement.value = value;
     }
 
+    // TODO: Do we really need all of this events
```
denisogolubev, line 62: And it will be also good to change `@HostListener` to the  global event handlers because they better optimized and do not affect the performance as host listeners do.
https://angular.love/simple-angular-context-help-component-or-how-global-event-listener-can-affect-your-performance

@zverbeta if you ok with such fix, add plz this to you TODO annotation.

denisogolubev (approved):

PR #2258, opened 2024-12-11
feat: Annotations from brace editor

denisogolubev:
- Add possibility to get actual annotations from brace editor
- allow to use `console.error()` in lint config

### DIFF
--- a/packages/components/ace-editor/xm-ace-editor.directive.ts
+++ b/packages/components/ace-editor/xm-ace-editor.directive.ts
@@ -29,6 +29,7 @@ export class XmAceEditorDirective<O = unknown> implements OnDestroy {
     public editor: Editor;
     public oldText: string;
     @Output() public textChanged: EventEmitter<string> = new EventEmitter<string>();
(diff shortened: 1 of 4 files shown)


packages/components/ace-editor/xm-ace-editor.directive.ts
```diff
@@ -153,4 +157,11 @@ export class XmAceEditorDirective<O = unknown> implements OnDestroy {
         this.oldText = newVal;
     }
 
+    private updateAnnotations(): void {
+        setTimeout(() => {
+            const annotations: any[] = this.editor.getSession().getAnnotations();
+            this.annotationsChange.emit(annotations);
+        }, 250);
```
oscaryald, line 164: 250 - might be move to const?

oscaryald (approved):

packages/components/ace-editor/xm-ace-editor.directive.ts
denisogolubev: True) No magic numbers) Fixed)
Thanks!

PR #2261, opened 2024-12-17
feat(sticky): recalculate sticky when column selection emit changes

Sumragen:


packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
```diff
@@ -89,7 +89,7 @@ export class XmTableSettingStore implements OnDestroy {
     }
 }
 
-@Injectable()
+@Injectable({providedIn: 'root'})
```
denisogolubev, line 92: Is that ok for the case when there are a several tables on the same page?

x3mlestat (approved):

packages/components/table/controllers/config/xm-table-columns-setting-storage.service.ts
Sumragen: Valid point. I've checked that scenario and it looks fine, column settings display correct set of columns and store it appropriately.

denisogolubev (approved):

PR #2271, opened 2025-01-17
add over mode to right sidebar

mak7iliano:


sergeysenja1992 (approved):

packages/components/sidebar-right/xm-sidebar-right.module.ts
```diff
@@ -141,7 +145,16 @@ export class XmSidebarRightComponent implements OnInit, OnDestroy {
 
     private changeMainElementMarginBy(width: string): void {
         const main: HTMLElement = document.getElementById('main');
```
zverbeta, line 147: You should send an event throu Event Broadcast service, and listen to this event in the main component

Event data shouldn't contain any margin value, it should depend on menu mode

zverbeta (changes requested):

packages/components/sidebar-right/xm-sidebar-right.module.ts
mak7iliano: this is old realisation, i have just added class toggling when mode is 'over'

zverbeta: I understand, but could we change the old pattern? I mean it could be better for understanding in the future

zverbeta (approved):

packages/components/sidebar-right/xm-sidebar-right.module.ts
mak7iliano: changed through the event

major13ua (approved):

PR #2284, opened 2025-02-13
Create websocket service

darlin200:


mak7iliano (approved):

packages/websockets/websocket.service.ts
```diff
@@ -0,0 +1,88 @@
...
+        };
+
+        this.socket.onmessage = (event) => {
+            console.warn('[XmWebSocket] Message received:', event.data);
+            this.messageSubject.next(event.data);
+        };
+
+        this.socket.onclose = (event) => {
+            console.warn(`[XmWebSocket] Disconnected. Code: ${event.code}, Reason: ${event.reason}`);
+            this.tryReconnect(typeKey);
```
sergeysenja1992, line 44: first try have to be with zero delay, second and nexts better to have progress delay like 
reconnectAttempts*delay

packages/websockets/websocket.service.ts
```diff
@@ -0,0 +1,88 @@
...
+            return this.messageSubject.asObservable();
+        }
+
+        const protocol = location.protocol === 'https:' ? 'wss' : 'ws';
+        const url = `${protocol}://${location.host}/websocket/connect?typeKey=${typeKey}&token=${token}`;
+
+        this.socket = new WebSocket(url);
+
+        this.socket.onopen = () => {
+            console.warn('[XmWebSocket] Connected.');
```
sergeysenja1992, line 33: reset reconnectAttempts when connected

packages/websockets/websocket.service.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Observable, Subject } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
+export class XmSocketService implements OnDestroy {
+    private tokenService: XmAuthenticationStoreService = inject(XmAuthenticationStoreService);
+    private socket: WebSocket | null = null;
+    private messageSubject = new Subject<string>();
+
+    private reconnectAttempts = 0;
+    private readonly maxReconnectAttempts = 5;
```
sergeysenja1992, line 12: why only 5?
In many cases it can be infinite 
Or have to be configurable

packages/websockets/websocket.service.ts
```diff
@@ -0,0 +1,88 @@
...
+    private readonly reconnectDelay = 3000;
+
+    public connect(typeKey: string): Observable<string> {
```
sergeysenja1992, line 15: we can have only one websocket? O_o
what will happens when I call connect with tow different typeKey?

packages/websockets/websocket.service.ts
darlin200: Will fix it next Commit

packages/websockets/websocket.service.ts
darlin200: It have to be configurable, right now is 5 cause it`s basic initialization. In next commit would be 30 (just in case why not)

packages/websockets/websocket.service.ts
darlin200: My plan was a basic commit but inizialization with multiple websockets is a plan. I already have. Next commit publish it

packages/websockets/websocket.service.ts
```diff
@@ -0,0 +1,111 @@
...
+import { XmAuthenticationStoreService } from '@xm-ngx/core/auth';
+import { Observable, Subject } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
+export class XmSocketService implements OnDestroy {
+    private sockets: Map<string, WebSocket> = new Map();
+    private messageSubject = new Subject<{ typeKey: string; message: string }>();
+
+    private reconnectAttempts = new Map<string, number>();
+    private readonly maxReconnectAttempts = Infinity;
```
sergeysenja1992, line 11: have to be configurable

PR #2317, opened 2025-04-07
feat: dashboard transfer

ivanperesadenko:
Dashboards Transfer is a functionality that allows transferring dashboards or roles from one environment to another. Currently, this functionality only allows the creation of dashboards and roles on the environments where we want to transfer them, but in the future, there will also be the possibility to update the existing ones

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "xm-webapp",
-  "version": "7.0.82",
+  "version": "7.0.83",
   "release": "6.0.0",
   "private": true,
   "description": "Xm-webapp",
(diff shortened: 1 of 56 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "xm-webapp",
-  "version": "7.0.71",
+  "version": "7.0.72",
```
major13ua, line 3: Please explain
- how environments are configured?
- how login to target environment is performed?
- do it allow to transfer configuration between tenants?
- it is one time transfer, or upsert mode supported?

major13ua (changes requested):

package.json
ivanperesadenko: Please note: its **MVP**

1. You can add an array of environment strings to the configuration object, and you will see your environments in the initial phase of autocomplete
2. Right now, there is no login functionality on the target environment. You need to do it manually and use a token. In the future, we will develop this functionality.
3. Technically, yes. You can be in the tenant A environment, manually log in to the tenant B environment, and use the domain and token to transfer whichever dashboards you want. For now, "transfer" specifically means creating only those that do not exist in the target environment.
4. Upsert mode is currently not supported—only creation is available. In subsequent development iterations, there are plans to develop functionality that allows determining whether a dashboard already exists in the target environment in order to avoid creating a duplicate and instead update the existing one

denisogolubev (approved):

x3mlestat (approved):

andreyminaev96 (approved):

major13ua (approved):

PR #2329, opened 2025-04-29
Adding example of keycloack call proxy pattern

major13ua:


sergeysenja1992 (changes requested):
pls use local.proxy.conf.js

PR #2334, opened 2025-05-12
feat(fab-button): Add fab-buttons feature

ivanperesadenko:
Fab buttons with controller support

### DIFF
--- a/packages/components/buttons/components/button-base.ts
+++ b/packages/components/buttons/components/button-base.ts
@@ -1,0 +1,33 @@
+import { Directive, inject, Input, OnDestroy, OnInit } from '@angular/core';
+import { injectByKey, XM_DYNAMIC_COMPONENT_CONFIG } from '@xm-ngx/dynamic';
+import { isObservable, Observable, of } from 'rxjs';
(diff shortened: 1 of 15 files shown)


packages/components/buttons/components/button-base.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    private readonly controller: ControllerBaseInterface = injectByKey<ControllerBaseInterface>(this.config?.controller?.key, { optional: true });
+
+    public ngOnInit(): void {
+        this.loading$ = this.controller?.loading$ || of(false);
+        this.disabled$ = this.controller?.disabled$ || of(false);
+    }
+
+    public executeControllerMethod(): void {
+        const executedMethod = this.controller[this.config.controller.method]();
```
Olensavchenko, line 24: We need an optional chaining for config

denisogolubev: Probably, it's even better to wrap this logic via `try catch`

denisogolubev (approved):

packages/components/buttons/components/button-base.ts
ivanperesadenko: done

Olensavchenko (approved):

oscaryald (approved):

PR #2342, opened 2025-06-02
Close sidebar by clicking out of the sidebar

darlin200:


packages/components/sidebar-right/xm-sidebar-right.module.ts
```diff
@@ -83,13 +83,24 @@ export class XmSidebarRightComponent implements OnInit, OnDestroy {
         }
     }
 
+    @HostListener('document:click', ['$event'])
+    public onDocumentClick(event: MouseEvent): void {
```
denisogolubev, line 87: Please, register this event via RxJS and only if in public UI config provided a flag for this.
```ts
import { XmUIConfig, XmUiConfigService } from '@xm-ngx/core/config';

private uiConfigService: XmUiConfigService<XmMainConfig> = inject(XmUiConfigService);

this.uiConfigService.config$().pipe(
            takeUntilOnDestroy(this),
            switchMap((config: XmMainConfig) => {
               // add boolean property isOutsideClickHideMenu in public UI config inside sidebar configuration.
                const { isOutsideClickHideMenu } = config?.sidebar || {};
                if (!isOutsideClickHideMenu) {
                    return of(null);
                }
                return fromEvent(window.document, 'click').pipe(
                    tap((event: MouseEvent) => {
                        const clickedInsideSidebar = this.elementRef.nativeElement.contains(event.target);
                        const clickedOnResizer = this.resizerElement?.nativeElement.contains(event.target);

                        if (!clickedInsideSidebar && !clickedOnResizer) {
                            this.remove();
                        }
                    })
                );
            })
        ).subscribe();
```

major13ua (approved):

PR #2357, opened 2025-06-19
add new service, DF-999

mak7iliano:


major13ua (approved):

packages/account/context/context.service.ts
```diff
@@ -0,0 +1,30 @@
+import { Injectable } from '@angular/core';
+import { XmUserService } from "@xm-ngx/core/user";
```
salatsynskahv, line 2: Please, change to single quotation mark, it affect lint build

PR #2386, opened 2025-08-13
Added dashboard configuration for not showing if session is not active

darlin200:
With logout for 0.5 seconds we have block of our dashboard. 
image

To fix it i`ve added session and logout event subscription for validate block

### DIFF
--- a/packages/dashboard/src/components/dashboard/dashboard.component.ts
+++ b/packages/dashboard/src/components/dashboard/dashboard.component.ts
@@ -55,14 +55,7 @@ export class DashboardComponent extends DashboardBase implements OnInit, OnDestr
 
     public ngOnInit(): void {
         this.xmEntitySpecWrapperService.spec().then((spec) => this.spec = spec);
-        this.eventManager.listenTo('USER-LOGOUT-INIT')
-            .pipe(takeUntilOnDestroy(this))
-            .subscribe(() => this.loggingOut$.next(true));
-      
-        merge(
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/components/dashboard/dashboard.component.ts
```diff
@@ -45,7 +55,14 @@ export class DashboardComponent extends DashboardBase implements OnInit, OnDestr
 
     public ngOnInit(): void {
         this.xmEntitySpecWrapperService.spec().then((spec) => this.spec = spec);
-
+        this.eventManager.listenTo('USER-LOGOUT-INIT')
```
denisogolubev, line 58: Can you please create a method for this and in onInit call only function? Just to avoid to expand code inside lifecycle hooks.
Like:
```ts
public ngOnInit(): void {
        // ...
        this.observeSessionEvents();
        // ...
    }
    
    private observeSessionEvents(): void {
        this.eventManager.listenTo('USER-LOGOUT-INIT')
            .pipe(takeUntilOnDestroy(this))
            .subscribe(() => this.loggingOut$.next(true));

        merge(
            this.eventManager.listenTo('USER-LOGOUT'),
            this.eventManager.listenTo('authenticationSuccess'),
        ).pipe(takeUntilOnDestroy(this)).subscribe(() => this.loggingOut$.next(false));
    }
```

denisogolubev (approved):

major13ua (approved):

PR #2552, opened 2026-04-15
role condition for actions

x3mlestat:


vk-smartflex (approved):

packages/components/table/components/xm-table-header.component.ts
```diff
@@ -29,8 +31,10 @@ import { NgIf } from '@angular/common';
             class="d-flex align-items-center flex-grow-1 justify-content-end"
         >
             <ng-content select="[refreshButton]"></ng-content>
-            <xm-table-actions-buttons class="d-flex align-items-center" [config]="config.actions">
-            </xm-table-actions-buttons>
+            <ng-container *xmCondition="config.editConditionActions; arguments {userAuthorities: userAuthorities}">
```
denisogolubev, line 34: It looks like specific business logic implementation in a component that should know nothing about it.
It is a container that displays every action in a table. The table supports multiple actions. editConditionActions wraps all actions, so one rule applies to all actions.
It's way better to make it possible to add a condition to the specific action, which allows you to configure different conditions for different actions.

denisogolubev (changes requested):

packages/components/table/components/xm-table-header.component.ts
x3mlestat: thanks, make sense, closed

PR #2555, opened 2026-04-16
possibility for search in nested by property

x3mlestat:


vk-smartflex (approved):

packages/operators/src/key-predicate.ts
```diff
@@ -1,10 +1,11 @@
 import { Primitive } from '@xm-ngx/interfaces';
-import { get, isArray, isPlainObject, findIndex } from 'lodash';
+import { get, isArray, isPlainObject, findIndex, has } from 'lodash';
 
 export interface NestedKeyFilters {
     [name: string]: {
         prop: string;
         predicate: Record<string, unknown>;
+        hasProperty?: string;
```
denisogolubev: Prefixes like `is, has, can, should, was, will, did, allows, supports, requires` must be used for the Boolean variables and functions. You have a string, so it's better to use another appropriate name for this property.

packages/operators/src/key-predicate.ts
```diff
@@ -24,8 +25,17 @@ export function searchNestedByPredicate(data: unknown[], keyFilters?: NestedKeyF
...
 
-        for (const { prop, predicate } of filters) {
-            const searchIndex = findIndex(stack, predicate);
+        for (const { prop, predicate, hasProperty } of filters) {
+            const searchIndex = findIndex(stack, (item) => {
+                const predicateMatch = findIndex([item], predicate) === 0;
+                if (!predicateMatch) {
+                    return false;
+                }
+                if (hasProperty) {
```
denisogolubev: It's not ok to do it this way because you do not follow SRP (Single Responsibility Principle) in this case.

packages/operators/src/key-predicate.ts
```diff
@@ -24,8 +25,17 @@ export function searchNestedByPredicate(data: unknown[], keyFilters?: NestedKeyF
 
         let stack = data;
 
-        for (const { prop, predicate } of filters) {
-            const searchIndex = findIndex(stack, predicate);
+        for (const { prop, predicate, hasProperty } of filters) {
+            const searchIndex = findIndex(stack, (item) => {
+                const predicateMatch = findIndex([item], predicate) === 0;
```
denisogolubev: Creating a new array [item] for every single element in the loop just to run findIndex on it is a waste of memory and CPU cycles. The original code used findIndex(stack, predicate) directly, which is highly optimized.

denisogolubev (changes requested):

denisogolubev:
Also, this function has `.spec.`. Is it still working?

packages/operators/src/key-predicate.ts
x3mlestat: thanks, refactored

packages/operators/src/key-predicate.ts
x3mlestat: thanks, refactored

packages/operators/src/key-predicate.ts
x3mlestat: thanks, refactored

packages/components/key-filter/controllers/filter-by-property-path.controller.ts
```diff
@@ -0,0 +1,21 @@
+import { Injectable } from '@angular/core';
+import { XmKeyFilterConfig, XmKeyFilterValue } from '@xm-ngx/components/key-filter';
```
denisogolubev: This import fails the build process. Use `../..`

packages/components/key-filter/controllers/filter-by-property-path.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+import { KeyFilterController } from '@xm-ngx/components/key-filter/key-filter.model';
```
denisogolubev: This import fails the build process. Use `../..`

packages/components/key-filter/controllers/filter-by-property-path.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+import { has } from 'lodash';
+
+@Injectable()
+export class FilterByPropertyPathController implements KeyFilterController{
```
denisogolubev: Run lint. It looks like here can be the issue

packages/components/key-filter/key-filter.component.ts
```diff
@@ -1,11 +1,13 @@
 import { ChangeDetectionStrategy, Component, Input, OnChanges } from '@angular/core';
-import { XmDynamicModule, XmDynamicPresentation } from '@xm-ngx/dynamic';
+import { KeyFilterController } from '@xm-ngx/components/key-filter/key-filter.model';
```
denisogolubev: This import fails the build process. Use `../..`

packages/components/key-filter/key-filter.model.ts
```diff
@@ -0,0 +1,5 @@
+import { XmKeyFilterConfig, XmKeyFilterValue } from '@xm-ngx/components/key-filter/key-filter.component';
```
denisogolubev: This import fails the build process. Use `../..`.
Also, you can move these interfaces from the component to this new model file.

packages/components/key-filter/controllers/filter-by-property-path.controller.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { TestBed } from '@angular/core/testing';
+import { FilterByPropertyPathController } from './filter-by-property-path.controller';
+import { XmKeyFilterConfig, XmKeyFilterValue } from '@xm-ngx/components/key-filter';
```
denisogolubev: Don't forget to change the import here after moving these interfaces to the model file

denisogolubev (changes requested):

packages/components/key-filter/controllers/filter-by-property-path.controller.ts
x3mlestat: fixed, thanks

packages/components/key-filter/controllers/filter-by-property-path.controller.ts
x3mlestat: fixed, thanks

packages/components/key-filter/controllers/filter-by-property-path.controller.ts
x3mlestat: fixed, lint is ok, thanks

packages/components/key-filter/key-filter.component.ts
x3mlestat: fixed, thanks

packages/components/key-filter/key-filter.model.ts
x3mlestat: fixed, thanks

packages/components/key-filter/controllers/filter-by-property-path.controller.spec.ts
x3mlestat: fixed, thanks

denisogolubev (approved):

Olensavchenko (approved):

grankovsky (approved):

PR #2567, opened 2026-05-04
feat(datetime-control): add min date in config for datetime-control

vk-smartflex:
## Summary by CodeRabbit

* **New Features**
  * Added ability to set a minimum date constraint for date/time fields, preventing selection of earlier dates.

* **Chores**
  * Package version bumped to 9.0.88.

### DIFF
--- a/packages/components/date/xm-datetime-control.ts
+++ b/packages/components/date/xm-datetime-control.ts
@@ -511,7 +511,20 @@ export class XmDateTimeControlComponent extends NgModelWrapper<XmDateTimeControl
     public messageErrors = inject<XmControlErrorsTranslates>(XM_CONTROL_ERRORS_TRANSLATES);
 
     @Input() public pickerFilter: XmDateTimePickerFilter;
-    @Input() public config: XmDateTimeControlConfig;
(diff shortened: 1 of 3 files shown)


packages/components/date/xm-datetime-control.ts
```diff
@@ -522,6 +528,10 @@ export class XmDateTimeControlComponent extends NgModelWrapper<XmDateTimeControl
             const value = this.value || interpolate(this.config.initValue, null);
             this.change(value);
         }
+
+        if (this.config?.minDate) {
```
denisogolubev: It's not safe to use `@Input` property in a constructor due to lifecycle execution.
Way better to use a setter for the `this.config` or computed Signal.

denisogolubev (changes requested):

packages/components/date/xm-datetime-control.ts
vk-smartflex: changed to setter

denisogolubev (approved):

Olensavchenko (approved):

grankovsky (approved):

PR #2577, opened 2026-05-15
feat(dynamic-dialog): add passing width and minWidth from config

vk-smartflex:
## Summary by CodeRabbit

* **New Features**
  * Dialogs now accept and apply caller-provided configuration options (e.g., sizing) when opened.
  * Added a key-filter controller that ensures non-array values are wrapped as single-element arrays; it is exported and made available to the components registry.

* **Chores**
  * Bumped package version to 9.0.93.

[Review Change Stack](https://app.coderabbit.ai/change-stack/xm-online/xm-webapp/pull/2577?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/components/dynamic/dynamic-dialog.ts
+++ b/packages/components/dynamic/dynamic-dialog.ts
@@ -15,7 +15,6 @@ export class DynamicDialog {
 
     public async createAsync<T extends XmDynamicPresentation, R>(selector: string, value: unknown, options: unknown): Promise<MatDialogRef<T, R>> {
         const matDialogRef = await this.getDialogRef<T, R>(selector, options);
-        debugger;
         matDialogRef.componentInstance.value = value;
         matDialogRef.componentInstance.options = options;
         return matDialogRef;
--- a/packages/components/dynamic/dynamic-dialog.ts
+++ b/packages/components/dynamic/dynamic-dialog.ts
@@ -14,19 +14,23 @@ export class DynamicDialog {
     }
 
     public async createAsync<T extends XmDynamicPresentation, R>(selector: string, value: unknown, options: unknown): Promise<MatDialogRef<T, R>> {
-        const matDialogRef = await this.getDialogRef<T, R>(selector);
+        const matDialogRef = await this.getDialogRef<T, R>(selector, options);
+        debugger;
         matDialogRef.componentInstance.value = value;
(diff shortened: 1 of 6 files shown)


packages/components/dynamic/dynamic-dialog.ts
```diff
@@ -14,19 +14,22 @@ export class DynamicDialog {
...
     public async createAsync<T extends XmDynamicPresentation, R>(selector: string, value: unknown, options: unknown): Promise<MatDialogRef<T, R>> {
-        const matDialogRef = await this.getDialogRef<T, R>(selector);
+        const matDialogRef = await this.getDialogRef<T, R>(selector, options);
         matDialogRef.componentInstance.value = value;
         matDialogRef.componentInstance.options = options;
         return matDialogRef;
     }
 
-    protected async getDialogRef<T, R>(selector: string): Promise<MatDialogRef<T, R>> {
+    protected async getDialogRef<T, R>(selector: string, options: unknown): Promise<MatDialogRef<T, R>> {
```
denisogolubev: The options should not be unknown. Use `DialogConfig` interface. Pass all options into `open` method.
Then no need to add another parameters in the future, because all possible will be supported.

packages/components/dynamic/dynamic-dialog.ts
```diff
@@ -14,19 +14,22 @@ export class DynamicDialog {
...
         const dialogComponent = await this.dynamicComponents.find<T>(selector, this.injector);
+        const {width, minWidth} = options as {width?: string, minWidth?: string};
```
denisogolubev: Does it work correctly without fallback?

denisogolubev (changes requested):

packages/components/dynamic/dynamic-dialog.ts
vk-smartflex: yes, it takes default one

packages/components/dynamic/dynamic-dialog.ts
vk-smartflex: done

denisogolubev (approved):

grankovsky (approved):

salatsynskahv (approved):

PR #2638, opened 2026-08-05
feat(multiselect): view value

Sumragen:
Fix 1: selectAll() now passes this.items.map(item => item.valueKey) instead of raw config.enum objects — fixes "Check All" not checking individual items
Fix 2: In syncControl(), moved this.items = this.toView(...) before selectAll() — fixes defaultSelectedAll: true being broken
Fix 3: Added selectedItemTitle getter and updated the trigger template to show the group name instead of the raw ID (e.g. "MyPerfectGroupName" instead of "2073134")

### DIFF
--- a/packages/components/multi-select/multi-select.component.ts
+++ b/packages/components/multi-select/multi-select.component.ts
@@ -60,7 +60,7 @@ type XmMultiSelectItemOrString = XmMultiSelectItem | string;
                     </ng-container>
 
                     <ng-template #someSelected>
-                        {{ (selectedLength ? selectedValues[0] : '') | translate }}
+                        {{ (selectedLength ? selectedItemTitle : '') | translate }}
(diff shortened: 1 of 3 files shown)


grankovsky (approved):

packages/components/multi-select/multi-select.component.ts
```diff
@@ -142,6 +142,11 @@ export class XmMultiSelectControlComponent extends NgFormAccessor<string[]> impl
         return this.selectedValues?.length;
     }
 
+    public get selectedItemTitle(): string {
```
denisogolubev, line 145: It's better to use a pipe or computed Signal here because a getter in a component with Default ChangeDetection is known to be a performance issue.

Olensavchenko (approved):

PR #2650, opened 2026-08-21
feat(enum-control) support multiple selected values

salatsynskahv:
## Summary by CodeRabbit

* **New Features**
  * Enum controls now support selecting multiple values.
  * Selected values remain visible, with the first selection shown alongside a count of additional selections.
  * Added an option to clear the current selection.

* **Bug Fixes**
  * Improved handling and display of selected enum options.

* **Chores**
  * Updated the package version to 9.1.11.

### DIFF
--- a/packages/components/enum/control/xm-enum-control.spec.ts
+++ b/packages/components/enum/control/xm-enum-control.spec.ts
@@ -86,4 +86,23 @@ describe('XmEnumControlComponent', () => {
         expect(component.value).toBe(value);
         expect(component.itemsMap[String(value)].value).toBe(value);
     });
+
+    it('multiple values should keep the selected values and resolve the first option', () => {
+        const value = [1, 2];
+        component.config = {
+            dataQa: '',
+            multiple: true,
+            items: [
+                { value: 1, title: 'First' },
+                { value: 2, title: 'Second' },
+            ],
+        };
+        component.value = value;
+        fixture.detectChanges();
+
+        expect(component.value).toEqual(value);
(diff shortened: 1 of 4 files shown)


x3mlestat (approved):

denisogolubev (approved):

grankovsky (approved):

packages/components/enum/control/xm-enum-control.spec.ts
```diff
@@ -86,4 +86,23 @@ describe('XmEnumControlComponent', () => {
...
+        const value = [1, 2];
+        component.config = {
+            dataQa: '',
+            multiple: true,
+            items: [
+                { value: 1, title: 'First' },
+                { value: 2, title: 'Second' },
+            ],
+        };
+        component.value = value;
```
dakasumov, line 100: We should test this via jsdom / happydom (which are included with Jest) rather than by checking component instances directly. Testing through component instances or internal methods means we are testing detached implementation details rather than actual behavior. Such tests can pass successfully even if the element does not actually exist in the DOM; the input is disabled or inaccessible to the user; event handlers (like click, change, etc.) are not properly bound to the DOM element due to markup or props issues and etc.
Using jsdom allows us to simulate real user interactions (finding the element, dispatching events, and checking the UI response). This brings our unit tests closer to how the software is actually used and helps catch potential UI/integration bugs early, preventing false-positive test results.

I recommend to watch this video by decoded FrontEnd: 
https://www.youtube.com/watch?v=BTEx2X_8b-U

